PR #2, opened 2019-10-19
Johnathan/signup

JohnathanZhou:
## Feature Name: Signup
For this feature, there were a couple main components:
- Sign up screen: basic frontend for form submission. We also used `react-native-navigation` to have this screen go to Homescreen on successful sign up (this is up in the air still depending on what we want to do with Login page).
- Form validation: validating that all the form inputs were good and checking the Airtable for duplicate phone numbers (our way of keeping track of users). We used validate.js and an async wrapper for the Airtable API call to do this 
- User persistence: we generate a user token upon successful sign up and store it in AsyncStorage, which we then check every time we load the app to decide whether or not we go to the authentication page or the home screen.

There were a few other little things like approving notifications upon sign up (Kyle's work) and modifying the homescreen to be a React component instead of function

## Future work

1. More elegant form validation. I set up the groundwork for using `onBlur` so that you have form validation on the go. I couldn't get it to work this sprint but it shouldn't be too much work.

2. Correct password validation. I created my JS RegEx pattern already but it wasn't completley working. Should be quick also

3. Styling
- Signup page styling
- Styled components going into separate files. Currently they're all stored in `style.container`

4. Putting validation components into a separate file. Currently they're all in the file because of a weird bug that I had (https://github.com/facebook/react-native/issues/4968).

## Asana Link
https://app.asana.com/0/1142438342256785/1145052574956250


.gitignore
```diff
@@ -1,3 +1,6 @@
+# Environment Variables
+env.sh
```
anniero98: Are these being used?

frontend/screens/SignUpScreen.js
```diff
@@ -0,0 +1,343 @@
...
+import React from "react";
+import { AsyncStorage, Button, Keyboard, StyleSheet, Text, TextInput, View } from "react-native";
+import validatejs from "validate.js";
+
+// I abstracted portions of the validation flow into these files 
+// but there's a weird bug "https://github.com/facebook/react-native/issues/4968"
+// so I put it all in this one file.
+// import { MonoText } from 'screens/signup/textfield.jsx'
+// import validation from  'screens/signup/validation'
+// import validate from  'screens/signup/validation_wrapper'
```
anniero98: Sounds good - could you start comments for future work with "TODO: @<github username>"? (Will communicate to this everyone going forward!) It's easier to find things we should go back and look at by searching in the codebase!

frontend/screens/signup/validation_wrapper.js
```diff
@@ -0,0 +1,35 @@
+import validatejs from 'validate.js'
+import validation from './validation'
+
```
anniero98: Nit: Add TODO comment to top of file

frontend/screens/signup/validation.js
```diff
@@ -0,0 +1,22 @@
+const validation = {
```
anniero98: Nit: Add TODO comment to top of file

frontend/screens/SignUpScreen.js
```diff
@@ -0,0 +1,343 @@
...
+          placeholder="Last Name"
+          onChangeText={(text) => this.setState({lastName:text})}
+          value={this.state.lastName}
+        />
+        <TextInput
+          style={styles.input}
+          placeholder="Phone Number"
+          onChangeText={(text) => this.setState({phoneNumber:text})}
+          value={this.state.phoneNumber}
+          // For future use to make forms even nicer
```
anniero98: Nit: Add TODO to beg. of comment

.gitignore
```diff
@@ -1,3 +1,6 @@
...
+
```
anniero98: Are these still being used?

.gitignore
```diff
@@ -57,5 +60,6 @@ typings/
 # dotenv environment variables file
 .env
 
+
```
anniero98: Nit: Revert changes to this file if above is unused ^

.gitignore
JohnathanZhou: Nope, will take out

PR #4, opened 2019-10-19
Login functionality

thumn:
**Functionality**
Users have the option of logging in by clicking on "Already have an account? Log in" button on the Sign Up screen. (In the future, we can discuss making either the login or sign up screens the main entry point).

Users log in with phone numbers in the format 1234567890, which is supported by the fact that the input field will only accept ten characters, and by default the keypad will be numeric.

Upon a successful Airtable lookup, we store the user session token in the Async storage in the form of first name + last name as retrieved from the Airtable. In the future, we can change this to return more information about the user or the record ID, which can point us to more information about the user.

When the user successfully logs in, we display their first and last name on the login page, clearing the form. This is a temporary change and just indicates that the login and user retrieval processes were successful.

**Testing**

- Logging in with a nonexistent phone number and/or incorrect password will display an error message on the login page.
- Logging in with a correct username and password will display the first and last name concatenated together on the login page.


thumn:
Squash and merge login functionality.
Major changes:
- SignUpScreen now contains a link to LoginScreen for users to log in.
- LoginScreen redirects to the HomeScreen upon successful login.
- LoginScreen was added as a route to the AppNavigator to allow SignUpScreen to access it.
- Other login screen functionality should be isolated from other screens.

PR #6, opened 2019-11-03
Tommy/stores_screen

tommypoa:
## Feature Name: Stores Screen
- Stores screen which consists of 2 components: MapView and BottomSheet
  - MapView shows markers on the maps that corresponds to the different stores. Clicking on it brings the user to the store's detailed screen
  - BottomSheet contains a list view of all the stores and their addresses and opening hours. Clicking on any store brings the user to the store's detailed screen

## Future Works
- MapView
  - Make it less laggy/buggy
  - Using google maps instead of default phone map app
  - Style maps
- Markers 
  - Markers rely on stores having longitude and latitudes -- to fill in the rest to render markers for all the other stores
  - Style markers icon
- Search bar
- Nearby and Favorites functionality


frontend/lib/common.js
```diff
@@ -6,10 +6,10 @@ import getEnvVars from '../environment';
 /**
  * Usage:
  * import { base } from "../lib/common.js"
- * BASE("Products")...
- **/
+ * base("Products")...
+ **/ 
 
-const { AIRTABLE_API_KEY, AIRTABLE_BASE_ID } = getEnvVars();
+const { AIRTABLE_API_KEY, AIRTABLE_BASE_ID, GOOGLE_MAPS_API_KEY } = getEnvVars();
```
anniero98: Do you want to remove this for this PR and leave a TODO comment instead? (Google maps key)

frontend/navigation/MainTabNavigator.js
```diff
@@ -1,20 +1,49 @@
...
-import { createBottomTabNavigator, createStackNavigator } from 'react-navigation';
-
-import TabBarIcon from '../components/TabBarIcon';
-import AnnouncementScreen from '../screens/AnnouncementScreen';
-import DetailedAnScreen from '../screens/DetailedAnnouncements';
-import HomeScreen from '../screens/HomeScreen';
-import ProductsDetailedScreen from '../screens/ProductsDetailedScreen';
-import ProductsScreen from '../screens/ProductsScreen';
-import SettingsScreen from '../screens/SettingsScreen';
+import React from "react";
```
anniero98: You know... I'm getting the feeling that you should install eslint/prettier.... because we have different formatting rules right now 😂

frontend/package.json
```diff
@@ -32,11 +32,18 @@
     "expo-constants": "^7.0.0",
     "expo-font": "^7.0.0",
     "expo-web-browser": "^7.0.0",
+    "google-map-react": "^1.1.5",
+    "install": "^0.13.0",
```
anniero98: I am curious what 'install' is 😮

frontend/screens/StoresDetailedScreen.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Image, View, Text } from 'react-native';
+
+import StoreInfo from '../components/StoreInfo';
+import { styles } from '../styles.js';
+
+class ProductsDetailedScreen extends React.Component {
```
anniero98: Class / export names are wrong!

frontend/screens/StoresDetailedScreen.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Image, View, Text } from 'react-native';
+
+import StoreInfo from '../components/StoreInfo';
+import { styles } from '../styles.js';
```
anniero98: Nitpick: remove `.js` *ahem eslint/prettier ; p*

frontend/screens/StoresScreen.js
```diff
@@ -0,0 +1,117 @@
...
+            renderContent={this.renderInner}
+            renderHeader={this.renderHeader}
+          />
+        </View>
+      </View>
+    );
+  }
+}
+
+function createStoreData(record) {
```
anniero98: extreme nit-pick, but why doesn't this function go at the top of the file? It's used in initialization so it would seem to make more sense readability-wise

(same for product screen)

frontend/package.json
anniero98: i still don't know but i uh believe in u

frontend/screens/StoresDetailedScreen.js
anniero98: I'll resolve this in master

frontend/package.json
anniero98: Ah I have realized I need to remove `google-map-react` from this commit. Will do in master

PR #8, opened 2019-11-20
Notifications: Signup/Login changes

anniero98:
## Notifications: Allow multiple devices
- Allow multiple devices per customer; i.e generate push token upon registration & upon login, and store multiple tokens in Airtable if new device.
- Refactored existing logic in `SignUpScreen`/`LoginScreen`. Pulled out a couple of functions into `authAirtable.js`.

## Future work
- Refactor validation into other file
- Refactor registerForPushNotificationsAsync to be non-state-dependent


screens/signup/authAirtable.js
```diff
@@ -0,0 +1,153 @@
...
+      })
+      .catch(err => reject(err));
+  });
+};
+
+// lookupCustomer searches for users based on their
+// phone numbers in the form (XXX) XXX-XXXX and checks against
+// a correct password. If the user is found, we return their first
+// and last name. Otherwise, we will display an error on the login screen.
+export const lookupCustomer = function async(phoneNumber, password) {
```
JohnathanZhou: This function can be written smaller but maybe something to do later so we can keep moving!

screens/signup/authAirtable.js
```diff
@@ -0,0 +1,153 @@
...
+          customerInfo = null;
+        }
+        // If no records exist, resolves with object of null fields
+        resolve(customerInfo);
+      })
+      .catch(err => reject(err));
+  });
+};
+
+export const updateCustomerPushTokens = function async(
```
JohnathanZhou: This function might want to be refactored too but same as earlier, should be good for now

screens/signup/authAirtable.js
FranCogMents: @kyleqhua , @JohnathanZhou makes a good point -- if you are only obtaining a maxRecord of 1 you can use `.firstPage` instead and check if it's not null and just extract the first element of that record list instead of iterating through the whole list

screens/signup/authAirtable.js
```diff
@@ -0,0 +1,153 @@
...
+        if (records.length !== 0) {
+          records.forEach(function id(record) {
+            customerInfo.custId = record.getId();
+            customerInfo.pushTokens = record.fields['Push Tokens'];
+            customerInfo.tokenNames = record.fields.token_names;
+          });
+        } else {
+          customerInfo = null;
+        }
+        // If no records exist, resolves with object of null fields
```
FranCogMents: @kyleqhua If no records exist, should the function resolve with a `null` `customerInfo` object or resolve with returning a `customerInfo` object with `null` fields?

screens/signup/notifications.js
```diff
@@ -0,0 +1,26 @@
+// import { Notifications } from 'expo';
```
FranCogMents: @anniero98 is this file being used in the project right now?

anniero98: No it is not! Will remove

screens/signup/authAirtable.js
anniero98: I realized it would be simpler to just return null object, but good catch - will update comments (:

.eslintrc.js
```diff
@@ -21,7 +21,8 @@ module.exports = {
     'constructor-super': 'warn',
     'valid-typeof': 'warn',
     'react/jsx-filename-extension': [1, { extensions: ['.js', '.jsx'] }],
-    'prettier/prettier': ['error']
+    'prettier/prettier': ['error'],
+    'react/destructuring-assignment': ['enabled', 'never'],
```
anniero98: Oops, forgot I did this. It's a bit annoying and I don't think we need to change this across the project right now, going to keep it

anniero98:
## Future work
As suggested by Franco, it's very possible we refactor everything to look like this instead - we l o v e Javascript
```
let customerInfo = null;
try {
    const customerInfo = await lookupCustomer(formattedPhoneNumber, this.state.password)
    const pushTokensRetVal = await updateCustomerPushTokens(customerInfo, this.state.token)
  } catch (e) {
    showError(e)
  }
```
However, this needs to get merged so I am going to merge it now (^:

PR #10, opened 2019-11-25
Fix breaking change in HomeScreen, refactor styling

anniero98:
# Work done
## styled-components
In preparation for future styling, broke out original `styles.js` into `styles/shared.js`, `styles/products.js`, `styles/stores.js`

## HomeScreen
Due to a change in configuration for Rewards (i.e using the Customer's stored point values to generate number of Rewards currently unlocked & available), deleted the Rewards table in Airtable. This is a breaking change, so removed all code relating to that. Will be cleaned up by @JohnathanZhou at a later point.

Did some light code cleanup with ESLint/Prettier as well.

# Future work
Will require migrating the current `styles` Stylesheets in various files to their own files in the new `styles` folder.


screens/HomeScreen.js
```diff
@@ -212,7 +158,7 @@ const styles = StyleSheet.create({
     fontWeight: 'bold',
```
tommypoa: Will this StyleSheet also be migrated into a separate style sheet for the homepage within the styles folder?

tommypoa: Just read your PR comment – futures work!

anniero98: Today's (or tomorrow's) work ~!

PR #11, opened 2019-11-29
Merge Products & Stores Screen

anniero98:
## Feature Name: Merge Products & Stores Screen
- Products and Stores (map) screens are merged
  - Store specific **products** are displayed in the BottomSheet
- First semblance of search bar exists!
  - Currently functions as a button to navigate to a list of stores
  - Clicking on any specific stores in the store list navigates back to the map / Stores screen
- MapView is now smooth with introduction of `onRegionChangeComplete` function
- All store screen helper functions are housed under `storeHelpers.js`

## Future Works
- Markers 
  - Style markers icon
- Search bar functionality
- Nearby and Favorites functionality


app.json
```diff
@@ -28,7 +28,8 @@
     },
     "android": {
       "package": "com.blueprint.dc",
-      "googleServicesFile": "./google-services.json"
+      "googleServicesFile": "./google-services.json",
```
anniero98: We aren't using this anymore, right? @tommypoa

components/Map.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from 'react';
```
anniero98: oops I thought I deleted this already

components/StoreProducts.js
```diff
@@ -0,0 +1,104 @@
...
+                })
+              }>
+              <Product product={product} />
+            </Button>
+          ))}
+        </ScrollView>
+        {/* Display vegetables available at this store */}
+        <View flexDirection="row">
+          <Title>Veggies</Title>
+          {/* TODO @tommypoa See all: pass current store as prop and show as Title if non-null prop */}
```
anniero98: (This TODO is done)

components/StoreProducts.js
```diff
@@ -0,0 +1,104 @@
...
+  render() {
+    const { navigation, products, store } = this.props;
+    return (
+      <View>
+        <StoreCard
+          store={store}
+          key={store.id}
+          callBack={() => this.detailedStoreTransition(store)}
+        />
+        {/* TODO @tommypoa can probably make this (e.g Fruits title + fruits products) a sub-component and re-use it in AllProducts */}
```
FranCogMents: @anniero98 TODO: Good idea if you have time since this pattern is used throughout the app

components/StoreProducts.js
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { View } from 'react-native';
+import { ScrollView } from 'react-native-gesture-handler';
+
+import { Button, Title } from '../styles/shared';
+import Product from './Product';
+import StoreCard from './StoreCard';
+
```
FranCogMents: @anniero98 since these categories are hard-coded and reused throughout this file -- let's make them global constants?

```
const CATEGORY  = {
FRUIT: 'fruit',
VEGETABLES: 'vegetables'
} 
const NAV = {
STORES_DETAILED = 'StoresDetailed',
PRODUCTS = 'products`
}
```

screens/AllProductsScreen.js
```diff
@@ -0,0 +1,121 @@
...
+
+import Product from '../components/Product';
+import BASE from '../lib/common';
+import { Button, ScrollCategory, H3, Title, Subtitle } from '../styles/shared';
+import { styles } from '../styles/products';
+
+const productsTable = BASE('Products').select({ view: 'Grid view' });
+let fullProducts;
+const categories = [
+  // Hard-coded for now -- should find a way to extract this information dynamically?
```
FranCogMents: @anniero98 

Unfortunately you can't query the possible options for multi-select fields -- but if you did wanted to get your categories dynamically you would need to make an additional table and do some record linking.  

One possible solution -- make a `Category` table with only 1 column `name` that has the category names   
  'All',
  'Cut Fruit & Packaged Products',
  'Fruit',
  'Vegetables',
  'Frozen & Dried'
And in your `products` table you make two columns -- one (you could name it `CategoryLink` will will link to the `category` 
Screen Shot 2019-11-30 at 11 27 35 PM

The other column, which you will actually name `category` (which will replace your current category column -- you can name it something else in the mean time) will be a of this type
Screen Shot 2019-11-30 at 11 30 34 PM
So now, if you want to just query all the categories you can now just query the Category table, and this will also maintain the `category` field in Products.

Although I see that you only handle Products with one category?

anniero98: Originally, we thought of having other categories such that they could be a multi-select (such as in-season produce, etc). Currently that idea has been scrapped, so we could indeed make it its own table & link the records!

There are few enough, though, that right now I think the global constants is a good approach. I'll note it down in a list of future works~! Wanted to get this PR merged in first - we'll refactor soon

components/StoreProducts.js
anniero98: We'll address this in another PR!

PR #13, opened 2019-11-30
Display customer points history

anniero98:
## Overview
This two-in-one PR adds the Points History Tab to the app and updates Announcement Styling. It pulls all of the user's transactions from Airtable sorted by date and displays them. Added basic styling to the Points History Tab as well as updated styling to the News Tab.

## Future
- [ ] Differentiate between rewards redeemed and points claimed.
- [ ] Divide transactions by months. 
- [ ] Add Icons to Transactions Cards.

https://app.asana.com/0/1142438342256785/1148207754507328/f
https://app.asana.com/0/1142438342256785/1146151404554647/f
https://app.asana.com/0/1142438342256785/1148207754507292/f


screens/historyHelpers.js
```diff
@@ -0,0 +1,41 @@
+import BASE from '../lib/common';
+
+const getCustomerTransactions = function async(userId) {
```
anniero98: This is really good! I realized while reviewing your code that there's a better way to interact with Airtable -

Refactor to use filterByFormula, and a sort function with the date! This way we don't have to iterate over _all_ transactions, and we only make a couple API calls versus other methods. And we can take advantage of Airtable built-in things like the sort function (":

Couple other code cleanup things:
- Use Promise.all  --> .then --> .map to create an array; use a helper function called 'createTransactionData' or something like that
- Here's an example from Tommy's PR (which I also refactored 😂)

```
      Promise.all(productRecords)
        .then(records => {
            const products = records.map(record => createProductData(record));
            resolve(products);
        })
    .catch(err => reject(err));
```

screens/PointsHistoryScreen.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+  }
+
+  async componentDidMount() {
+    const userId = await AsyncStorage.getItem('userId');
+    getCustomerTransactions(userId).then(transactions => {
+      this.setState({ transactions });
+    });
+
+    Font.loadAsync({
```
anniero98: I moved this here; not sure where it actually belongs but generally try to avoid having functions living + getting called on their own inside the class itself

screens/PointsHistoryScreen.js
```diff
@@ -0,0 +1,44 @@
...
+      Poppins: require('../assets/fonts/Poppins-Regular.ttf')
+    });
+  }
+
+  render() {
+    return (
+      <View>
+        <Text> RECENT TRANSACTIONS</Text>
+        <ScrollView>
+          {this.state.transactions.map(transaction => (
```
anniero98: Here, display recents first and add appropriate FE dividers etc as in Figma

screens/PointsHistoryScreen.js
```diff
@@ -0,0 +1,44 @@
...
+    super(props);
+    this.state = {
+      transactions: []
+    };
+  }
+
+  async componentDidMount() {
+    const userId = await AsyncStorage.getItem('userId');
+    getCustomerTransactions(userId).then(transactions => {
+      this.setState({ transactions });
```
anniero98: This was the fix -- when Promises resolve/return some value (for a .then -> .then case), you have to use the result in a .then call. Here we actually want to set the `transactions` object in state so that when we render using `this.state.transactions`, we'll actually have something in that object~

anniero98: You'll want to split out the three most recent transactions into a separate array (three being arbitrary and as specified in Asana; maybe call it 'recents' and store that in state as well. That way you can easily display it later ⬇️

As always, refer to Figma for the wireframes

screens/PointsHistoryScreen.js
```diff
@@ -0,0 +1,44 @@
...
+
+    Font.loadAsync({
+      Poppins: require('../assets/fonts/Poppins-Regular.ttf')
+    });
+  }
+
+  render() {
+    return (
+      <View>
+        <Text> RECENT TRANSACTIONS</Text>
```
anniero98: Super nitpick but this should go inside ScrollView, esp. after you split out transactions into the `recents` state object

screens/historyHelpers.js
```diff
@@ -0,0 +1,41 @@
+import BASE from '../lib/common';
+
```
anniero98: I also moved this function into this folder, because I anticipated needing more helper functions before realizing we could use `filterByFormula` .. for now I think it's okay though; our other screens are slowly moving functions out as well

components/Transactions.js
```diff
@@ -0,0 +1,33 @@
...
+  IconContainer,
+  MainText,
+  Overline
+} from '../styles/transactions';
+
+/**
+ * @prop
+ * */
+
+function Transactions(props) {
```
anniero98: I modified this component to be a function instead -- generally if we don't need state at all and only rely on props, it means we should likely just make this a function. Also because eslint yelled at me

components/Transactions.js
```diff
@@ -0,0 +1,33 @@
...
+  const { date, storeName, points } = props;
+  return (
+    <TouchableOpacity>
+      <Card>
+        <IconContainer />
+        <ContentContainer>
+          <Overline>
+            {date.toLocaleDateString()} @ {storeName}
```
anniero98: I added a 'Lookup' field in Airtable so that you could get 'storeName' back as expected, instead of a RECORD_ID (getting the value from a linked record field will always give you a record_id regardless of the primary key displayed in Airtable; so even though 'Store' shows the store names, when you get that value in code it's a record ID)

anniero98: I also changed all instances to be 'storeName' instead of 'storeId' for clarity

kyleqhua: Thats actually so much more efficient omg

components/Announcements.js
```diff
@@ -1,70 +1,50 @@
...
-        flexDirection: 'column'},
-    date: {
-        fontWeight: 'bold',
-    },
-    title: {
-        color: 'blue',
-    },
-
-});
+import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
```
anniero98: It would have been better to do Announcements in a separate PR, actually 😅PRs are meant to keep changes modularized! If you don't already know how to undo/move to a new branch, it's ok! A bit too late now since we aren't meeting in person for awhile -- sometime I'll teach you some fun git stuff to do it

screens/AnnouncementScreen.js
```diff
@@ -41,6 +43,12 @@ class AnnouncementScreen extends React.Component {
     };
   }
 
+  async componentDidMount() {
```
anniero98: This doesn't need to be `async`, I think it breaks if you do?

screens/AnnouncementScreen.js
```diff
@@ -41,6 +43,12 @@ class AnnouncementScreen extends React.Component {
...
+    Font.loadAsync({
+      Poppins: require('../assets/fonts/Poppins-Regular.ttf')
+    });
+  }
+
   render() {
     return (
       <View>
```
anniero98: Nitpick: We don't need the 'Events' tab at all anymore; check the 'Week 15' in Figma while you are styling please!

components/Announcements.js
kyleqhua: Oh rip Im sorry :(

screens/AnnouncementScreen.js
kyleqhua: You right

components/Announcements.js
anniero98: It's ok! Just something to keep in mind for the future (:

components/Transactions.js
anniero98: Airtable magic (^: woo

screens/historyHelpers.js
anniero98: Wow it really is much shorter now

screens/PointsHistoryScreen.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+  }
+
+  async componentDidMount() {
+    const userId = await AsyncStorage.getItem('userId');
+    getCustomerTransactions(userId).then(transactions => {
+      const recent = [];
+      recent.push(transactions.shift());
+      recent.push(transactions.shift());
+      recent.push(transactions.shift());
```
JohnathanZhou: Probably a little redundant to call `.shift()` 3x here, I'd recommend using the splice method! Shift is also memory intensive because it has to move the move every item in the array over

`const recent = transactions.splice(0, 3).reverse();` should achieve the same thing you want in!

screens/AnnouncementScreen.js
```diff
@@ -1,11 +1,19 @@
...
+  TopText,
+} from '../styles/announcements';
 const announceTable = BASE('Announcements').select({ view: 'Grid view' });
 const firstAnnouncements = [];
+
+Font.loadAsync({
+  Poppins: require('../assets/fonts/Poppins-Regular.ttf')
+});
+
 announceTable.eachPage(
```
JohnathanZhou: I'd recommend doing this inside `componentDidMount` instead. I don't remember the exact details but I don't believe it's good practice to have important code running here -- we want to keep it within the component lifecycle

PR #16, opened 2019-12-03
Tommy/products_refactor

tommypoa:
### Files touched and refactored
- StoreProducts.js
  - Created helper function renderProducts to address code duplication
- Moved ProductsDetailedScreen.js and ProductsScreen onto a products folder, similar to stores


components/StoreProducts.js
```diff
@@ -21,68 +21,42 @@ function filterVegetables(product) {
...
   detailedStoreTransition = store => {
     this.props.navigation.navigate('StoresDetailed', {
       currentStore: store
     });
   };
 
-  render() {
-    const { navigation, products, store } = this.props;
+  renderProducts = (filterType, productType) => {
+    const { navigation, store, products } = this.state;
```
anniero98: The reason this was changed to props is because as a component not at the top-level screen, React won't force it to re-render; even if we set the state here, it simply doesn't trigger any hook in the lifecycle methods etc. 

I realized this while implementing location, which is why the constructor & this.state dependencies were removed -- here, I think we can solve it by destructuring props in `render` and passing args directly to renderProducts (lot of args 😅 but it's ok)

PR #20, opened 2019-12-08
Johnathan/bottomtab hamburger

JohnathanZhou:
## PR Details
* Drawer navigation. We are no longer using the bottom tab navigation as of week 14 (from the Figma), so a drawer better serves our purpose. This PR revamps our navigation system and now maps all the routes in a drawer navigator instead of the tab navigator we had before
* Small frontend things. Added a green banner on the bottom of StoresScreen that takes you to `RewardsScreen` (homescreen in the code right now). Others include adding a hamburger that opens a drawer menu on each page, removing headers on all the pages, and a few small changes to the title text in announcements and history.

## Future
- [ ] Merging homescreen (which will also be renamed rewardsscreen in a different PR) with points history. This will probably be done after tabs are ready 
- [ ] Handling the points on the green banner on StoresScreen. I have some ideas on this but might need a bigger change in design for it to not load airtable data numerous times.
- [ ] Better header handling. Currently we don't use the headers anywhere (I don't like how it looks personally and some functionality wasn't working with header) but the current way of showing each page title isn't super standardized
-  [ ] Code cleanup. I went through this in hackathon in kind of one go, so a lot of stuff to tidy up still and clean up more. More commits to come on this soon probably! A few quick things I can think of right now: removing `MainTabNavigator.js`, removing comments on the `AppNavigator`, and a couple more. 

https://app.asana.com/0/1142438342256785/1151837300693449


screens/AnnouncementScreen.js
```diff
@@ -1,68 +1,64 @@
...
     super(props);
     this.state = {
-      announcements: firstAnnouncements
+      announcements: []
     };
   }
 
   async componentDidMount() {
-
+    announceTable.firstPage()
```
anniero98: Ah.. yes this is the only one that didn't get refactored for the API calls 😂

PR #23, opened 2020-02-08
[0] Development lifecycle improvements 

anniero98:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR

[//]: # "Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

### Updating env config
Chose babel-plugin-inline-dotenv over react-native-dotenv because this person's rationale seemed quite reasonable
https://stackoverflow.com/questions/57060316/environment-variables-in-babel-config-js

Placed a `.env.example` file in root - duplicate it to create `.env.development` and `.env.production` files. For now, the latter two files will be the same. Fill in your Airtable API key, and the IMG API Key from @JohnathanZhou. 
**Please make sure not to commit your old `environment.js` file after pulling from master**

### Airtable refactoring
- See Aivant's Airtable generator repo on Github https://github.com/aivantg/airtable-schema-generator/
!! Important: I haven't actually changed our Airtable calls to use this yet - I'm still figuring out what I want that to look like. That'll be in the PR after the restructuring PR #24

### Package updates
I resolved all the errors and warnings from `yarn` when running `yarn install`. This will be done automatically from now on.

#### Post-pull/merge/etc: yarn package management

Using package `yarnhook` to watch `package.json` and `yarn.lock` for changes (I think); basically it'll automatically trigger a `yarn install` to make sure your packages are up-to-date.

#### Pre-commit: Prettier and ESLint run before commit
`husky` enables easy usage of git hooks.

As the name suggests, `lint-staged` runs a series of scripts on only staged files

**From now on, contributors  _must_ have ESLint working with the VSCode setup. Why? Because we'll add back the pre-commit `lint-staged` script that runs `eslint --fix` and won't allow you to commit without fixing the errors.**

!! We will not allow you to make any contributions to the codebase without following ESLint now (^:

#### Aside: VSCode auto-save conflicts 
(_whack thing that was happening to me_)

`yarn add --dev eslint-config-prettier eslint-plugin-prettier`

In `.eslintrc.js`: 
```
	plugins: ['prettier'],
	extends: ['plugin:prettier/recommended']
```
** Make sure it's at the end of the 'extends' list, so it can override any existing configs. See https://prettier.io/docs/en/integrating-with-linters.html

In your VSCode `settings.json`, make sure these lines exist
```
	"editor.defaultFormatter": "esbenp.prettier-vscode",
    "[javascript]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode"
[...]

PR #25, opened 2020-02-09
Store search, EBT status and BottomSheet bug fix.

tommypoa:
# What's new in this PR
### Store Search
- Search bar functionality added for Store List view – store will show as long as search term is a substring of the store name
- Search currently only works with/filters by store names

### EBT status
- Added additional EBT field in store object and displayed in StoreCard
- Currently displayed as a Yes/No field

### BottomSheet bug fix
- Bug: Bottom Sheet might disappear if user swipes the bottom sheet up aggressively
- Increased upper limit from 200 - 500%, which should be sufficient to prevent the bug from being triggered by the user

# Next steps
- Change out EBT status from Yes/No to icons

CC: @wangannie @anniero98

### DIFF
--- a/components/StoreCard.js
+++ b/components/StoreCard.js
@@ -9,7 +9,7 @@ import { H3, InLineContainer, Subtitle, Title } from '../styles/shared';
  * */
 
 function StoreCard({ store, callBack }) {
-  const { name, hours, address, distance } = store;
+  const { name, hours, address, distance, ebt } = store;
   return (
     <View style={{ marginBottom: 10 }}>
       <TouchableOpacity onPress={callBack}>
@@ -23,6 +23,9 @@ function StoreCard({ store, callBack }) {
           <FontAwesome name="clock-o" size={12} />
           <H3>{hours}</H3>
         </InLineContainer>
+        <H3>
+          EBT: {ebt ? <Subtitle>yuh</Subtitle> : <Subtitle>nuu</Subtitle>}
+        </H3>
       </TouchableOpacity>
     </View>
   );
--- a/components/StoreCard.js
+++ b/components/StoreCard.js
@@ -23,9 +23,7 @@ function StoreCard({ store, callBack }) {
           <FontAwesome name="clock-o" size={12} />
           <H3>{hours}</H3>
         </InLineContainer>
-        <H3>
-          EBT: {ebt ? <Subtitle>yuh</Subtitle> : <Subtitle>nuu</Subtitle>}
-        </H3>
+        <H3>EBT: {ebt ? <Subtitle>Yes</Subtitle> : <Subtitle>No</Subtitle>}</H3>
       </TouchableOpacity>
     </View>
   );
--- a/screens/stores/StoreListScreen.js
+++ b/screens/stores/StoreListScreen.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ScrollView } from 'react-native-gesture-handler';
-import { SearchBar } from 'react-native-elements';
+import { SearchBar } from 'react-native-elements'; // @tommypoa: Create styled-component for this
 
 import StoreCard from '../../components/StoreCard';
 import { Title } from '../../styles/shared';
@@ -10,7 +10,12 @@ class StoreListScreen extends React.Component {
   constructor(props) {
     super(props);
     const { stores, navigation } = this.props.navigation.state.params;
-    this.state = { stores, navigation, search: '', searchStores: stores };
+    this.state = {
+      allStores: stores,
+      navigation,
+      searchStr: '',
+      filteredStores: stores
+    };
   }
 
   detailedStoreTransition = store => {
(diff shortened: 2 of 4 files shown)


components/StoreCard.js
```diff
@@ -23,6 +23,9 @@ function StoreCard({ store, callBack }) {
           <FontAwesome name="clock-o" size={12} />
           <H3>{hours}</H3>
         </InLineContainer>
+        <H3>
+          EBT: {ebt ? <Subtitle>yuh</Subtitle> : <Subtitle>nuu</Subtitle>}
```
anniero98: @tommypoa .....

screens/stores/StoreListScreen.js
```diff
@@ -9,7 +10,7 @@ class StoreListScreen extends React.Component {
   constructor(props) {
     super(props);
     const { stores, navigation } = this.props.navigation.state.params;
-    this.state = { stores, navigation };
+    this.state = { stores, navigation, search: '', searchStores: stores };
```
anniero98: Nit: Can we name these a little better? `stores` is probably better off as `allStores` since it's continuously being used in the `.filter`; `searchStores` --> `searchedStores`, or `filteredStores`, or anything else that captures what it is a little better

anniero98: Nit: Can we rename these variables? `stores` is probably better off as `allStores` since it's continuously being used in the `.filter`; `searchStores` --> `searchedStores`, or `filteredStores`, or anything else that captures what it is a little better

anniero98: Similarly, I feel like `searchString` or `searchStr` might be better than `search`

screens/stores/StoreListScreen.js
```diff
@@ -18,12 +19,33 @@ class StoreListScreen extends React.Component {
...
+    };
+  }
+
   render() {
+    const { search } = this.state;
+
     return (
       <StoreModal>
+        {/* Search bar */}
+        <SearchBar
```
anniero98: Nit: Should add a todo for styling, since SearchBar is from a library and not from `styled-components`

screens/stores/StoresScreen.js
```diff
@@ -9,19 +9,25 @@ import {
...
+
 import MapView, { Marker } from 'react-native-maps';
 import BottomSheet from 'reanimated-bottom-sheet';
-import Hamburger from '../../components/Hamburger'
+import Hamburger from '../../components/Hamburger';
 
 import StoreProducts from '../../components/StoreProducts';
 import { Subtitle } from '../../styles/shared';
-import { SearchBar, StoreModal, StoreModalBar, TopText } from '../../styles/stores';
+import {
```
anniero98: eslint? prettier? I approve (peep #23 for pre-commit hooks in our future 👀)

screens/stores/StoresScreen.js
```diff
@@ -239,7 +245,7 @@ export default class StoresScreen extends React.Component {
             initialSnap={1}
             enabledInnerScrolling
             enabledGestureInteraction
-            snapPoints={['200%', '45%', '10%']}
+            snapPoints={['500%', '45%', '10%']}
```
anniero98: Nit: You might want to add a comment here for clarity - it's unintuitive from reading the code that 500% is a significant number, or that changing it would cause a bug

screens/stores/StoreListScreen.js
tommypoa: Commented. Not sure how involved this ticket will be so am pushing it to the next sprint (also might be a good learning opp for new mem).

screens/stores/StoresScreen.js
tommypoa: Included comments with explanation - not sure if this borders on verbose but looks alright now.

screens/stores/StoreListScreen.js
```diff
@@ -18,12 +24,33 @@ class StoreListScreen extends React.Component {
...
+  }
+
   render() {
+    const { searchStr } = this.state;
+
     return (
       <StoreModal>
+        {/* Search bar */}
+        <SearchBar
+          placeholder="Type Here..."
```
thumn: Maybe replace placeholder with something more intuitive like "enter store name here" or "Store name"? But if you feel like everything else looks intuitive then feel free to keep :0

screens/stores/StoresScreen.js
```diff
@@ -9,19 +9,25 @@ import {
...
+  TopText
+} from '../../styles/stores';
 import { getProductData, getStoreData } from './storeHelpers';
 
-
-const {width} = Dimensions.get('window'); // full width
+const { width } = Dimensions.get('window'); // full width
 
 // TODO is this const necessary?
 const deltas = {
```
JohnathanZhou: I think it is necessary , don't see a reason to not use const here

screens/stores/StoreListScreen.js
wangannie: +1 Change placeholder text to "Search by Store Name" to reflect the Figma and make it explicit what exactly users can search by. It seems like a nit but this is very important for real usability!

PR #26, opened 2020-02-10
Created reusable base components

thumn:
**New Features**
- Imported Poppins Regular font
- Created components for headers, text, and buttons in `BaseComponents.js`

**How to Review**
Components are currently in a standalone file and aren't being referenced in any other files. Changing existing button or header tags to new shared component tags would be helpful in seeing how they work, or we can make another PR restyling one screen using the new components.

**Tests**
None explicitly -- just replaced a few of the components on the sign up form while creating these components. Upon starting up the app, everything should be the same as it was before I started this PR!

CC: @wangannie

### DIFF
--- a/App.js
+++ b/App.js
@@ -8,6 +8,9 @@ import { Ionicons } from '@expo/vector-icons';
 import AppNavigator from './navigation/AppNavigator';
 
 export default function App(props) {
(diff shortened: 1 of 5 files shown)


thumn:
IMG_1351

This is what it looks like!

PR #27, opened 2020-02-10
Navigation changes + news fix

JohnathanZhou:
## What's new
* Made it so that you can't swipe to open the drawer
* Added a top tab to the rewards screen that takes you back to the map screen
* Small fix on news screen so that the hamburger icon doesn't show but instead there's an X to close the screen and go to the most recent screen

## Todo
- [x] Drawer still a little buggy in that you can swipe it out, will look into
- [ ] Code tidying -- I created a whole new button for closing inside the news page but I can definitely combine it with the hamburger component and make it more reusable 
- [x] Add an icon that makes it clear that you click in the green area to go back to the stores screen (aka the ↑ that is in the Figma)

### DIFF
--- a/screens/AnnouncementScreen.js
+++ b/screens/AnnouncementScreen.js
@@ -2,7 +2,6 @@ import React from 'react';
 import { ScrollView, Text, ViewComponent, TouchableOpacity, Button, View } from 'react-native';
 
 import Announcements from '../components/Announcements';
-import Hamburger from '../components/Hamburger';
 import { TopText } from '../styles/announcements';
 import getAnnouncements from './AnnouncementHelper';
 
@@ -24,6 +23,7 @@ class AnnouncementScreen extends React.Component {
     // const {goBack} = this.props.navigation;
     return (
       <View>
+        {/* TODO @johnathanzhou make this into a new component */}
         <TouchableOpacity
               style={{
                   backgroundColor:'white',
--- a/screens/AnnouncementScreen.js
+++ b/screens/AnnouncementScreen.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ScrollView, Text, ViewComponent, TouchableOpacity, Button, View } from 'react-native';
+import { ScrollView, Text, TouchableOpacity, Button, View } from 'react-native';
 
 import Announcements from '../components/Announcements';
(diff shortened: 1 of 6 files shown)


anniero98:
"Created components for headers, text, and buttons in BaseComponents.js" not sure if duplicate work w/ #26 , or just copy/paste 😂

screens/AnnouncementScreen.js
```diff
@@ -21,9 +21,32 @@ class AnnouncementScreen extends React.Component {
   }
 
   render() {
+    // const {goBack} = this.props.navigation;
     return (
       <View>
-        <Hamburger navigation={this.props.navigation} />
+        <TouchableOpacity
```
tommypoa: Think there's potential for this button to be made into a (styled-)component given how verbose it looks in the current code + possibility of re-use in other screens if DC chooses to add additional screens in their sidebar.

screens/rewards/RewardsScreen.js
```diff
@@ -170,7 +175,13 @@ export default class RewardsScreen extends React.Component {
               onRefresh={this._onRefresh}
             />
           }>
-          <Text styles={{ fontSize: 24 }}> Healthy Rewards </Text>
+          <TouchableOpacity
```
tommypoa: Works fine but not an intuitive trigger from a user stand-point since there is no icon + the trigger area is actually quite large! According to our current design/Figma, it seems like there should be an arrow down button that triggers it - possible stream-lining with the News cross button that you created?

tommypoa: Also stretch: making the screen transition between screens smoother somehow

screens/rewards/RewardsScreen.js
```diff
@@ -208,6 +219,18 @@ const styles = StyleSheet.create({
     textAlign: 'center'
   },
   tabView: {
-    flex: 1
+    flex: 1,
```
tommypoa: Looks good, but as a todo I feel like we should create a separate style-sheet or components for this screen (as is the case for the other screens).

screens/AnnouncementScreen.js
JohnathanZhou: Added as a todo for now

PR #28, opened 2020-02-18
New reusable base components + updates from wangannie

thumn:
Same [PR](https://github.com/calblueprint/dccentralkitchen/pull/26) with updates from @wangannie for testing in StoreProducts and StoreCard and loading remaining Poppins styles


thumn:
@FranCogMents Saw your comment on the [previous PR](https://github.com/calblueprint/dccentralkitchen/pull/26)! This PR is basically the same as the original one I had opened, with some of @wangannie's updates.

In terms of reducing to components we actually need, I made these components based off Figma designs that @wangannie gave me, and I believe all of them are needed as of right now on various screens! So I'll just go ahead and keep the components in for now if that's ok :~)

App.js
```diff
@@ -8,6 +8,9 @@ import { Ionicons } from '@expo/vector-icons';
 import AppNavigator from './navigation/AppNavigator';
 
 export default function App(props) {
+  Font.loadAsync({
```
anniero98: I think we're actually already loading this in line 46 -- it seems to make sense to do so in the `loadResourcesAsync` function that was built-in with Expo

components/StoreCard.js
```diff
@@ -1,8 +1,8 @@
 import { FontAwesome, MaterialCommunityIcons } from '@expo/vector-icons';
 import React from 'react';
 import { TouchableOpacity, View } from 'react-native';
-
-import { H3, InLineContainer, Subtitle, Title } from '../styles/shared';
+import {Subhead, Body, Title,Caption} from '../components/BaseComponents';
+import { H3, InLineContainer, Subtitle } from '../styles/shared';
```
anniero98: Let's get rid of any of these that are unused now - linter should have complained

components/StoreProducts.js
```diff
@@ -4,7 +4,8 @@ import { View } from 'react-native';
 import { ScrollView } from 'react-native-gesture-handler';
 import Product from './Product';
 
-import { Button, Title } from '../styles/shared';
+import {TextButton} from '../components/BaseComponents';
+import { Button ,Title} from '../styles/shared';
```
anniero98: Same as above - let's get rid of anything unused with these changes.

Also, keep in mind that this means this needs to be merged _after_  #25, and the changes here may need to be updated for elements @tommypoa added (minimal IIRC)

App.js
anniero98: Also, just noticed but if we're adding `Poppins-Medium` and `Poppins-SemiBold` to the assets, it would make sense to add the load calls in this PR. Otherwise, we should get rid of them. @wangannie 

IMO we should add the `load` calls to `loadResourcesAsync` as well

components/BaseComponents.js
```diff
@@ -0,0 +1,108 @@
...
+  font-size: 14px;
+  line-height: 20px;
+  display: flex;
+  align-items: center;
+  text-align: center;
+  color: #008550;
+`;
+
+export const TextButtonPressed = styled(TextButton)`
+  color: #004e14;
```
FranCogMents: Do these color's have a corresponding names in Figma? And do these colors already exist in a globals styles variable file that we could import and use? 

For readability and reusability would be great if the colors would read something with a name

`export const TextButtonPressed = styled(TextButton)`
  color: ${primary-green};
`;
`

components/BaseComponents.js
```diff
@@ -0,0 +1,108 @@
...
+  font-style: normal;
+  font-weight: 600;
+  font-size: 14px;
+  line-height: 21px;
+  display: flex;
+  align-items: center;
+  text-align: center;
+  letter-spacing: 0.01;
+  text-transform: uppercase;
+  color: ${props => (props.filledButton ? '#ffffff' : '#000000')};
```
wangannie: Add something like this to every text style (or at least the ones with frequently varying colors like Body Caption and Title) but directly passing in the color name (once color names are defined) into the styled component with a default color if nothing is specified.
`color: ${props => (props.textColor ? props.textColor : 'text-primary')}`

cc @FranCogMents

App.js
thumn: Sounds good! :)

components/BaseComponents.js
thumn: Switching to camel case for the naming because that seems to be the norm!

components/BaseComponents.js
```diff
@@ -0,0 +1,108 @@
...
+  align-items: center;
+  text-align: center;
+  color: #008550;
+`;
+
+export const TextButtonPressed = styled(TextButton)`
+  color: #004e14;
+`;
+
+export const TextButtonDisabled = styled(TextButton)`
```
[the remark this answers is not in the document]
thumn: I considered doing this, but it blocked me from doing a ternary operation to select from more than two color options :(

components/BaseComponents.js
thumn: Thanks!!! Realized I was thinking about nested ternary operations the wrong way :~)

components/BaseComponents.js
```diff
@@ -0,0 +1,108 @@
...
+`;
+
+export const FilledButtonContainer = styled(ButtonContainer)`
+  width: 344px;
+  height: 40px;
+  background: #008550;
+  border-radius: 20px;
+`;
+
+export const FilledButtonContainerPressed = styled(FilledButtonContainer)`
```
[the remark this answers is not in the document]
thumn: I'm doing this in `FilledButtonContainer` but for some reason it looks like it's not getting parsed correctly? Do you or @FranCogMents know anything about why this is wrong :(

`${props =>
    props.pressed
      ? 'box-shadow: 0px 11px 15px rgba(0, 0, 0, 0.2), 0px 24px 38px rgba(0, 0, 0, 0.14);'
      : ''}`

The error is like "Failed to parse declaration "boxShadow: ...""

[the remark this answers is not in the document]
anniero98: @thumn Could you try using hex? Or have it referenced from `Colors.js` (':  (ref. 1951's code https://github.com/calblueprint/1951coffee/blob/8bfc356e31c347b9a30f1263665426316b3ddcd1/components/BaseComponents.tsx#L12)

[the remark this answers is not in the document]
thumn: It looks like the problem comes from having two separate values for box-shadow :( it seems to compile when I do only the first half. But also @wangannie we may not need the darker green/box-shadow effect? The touchable opacity component itself already has a press effect that we can rely on.

PR #29, opened 2020-02-23
Update Colors.js with all colors, add colors to StoreCard

anniero98:
- Updated `Colors.js` to have all colors from Figma so we can use them when styling
- Hotfix for missing import in `StoreCard.js`, also added colors to the text to test out passing color as props

### DIFF
--- a/assets/Colors.js
+++ b/assets/Colors.js
@@ -1,10 +1,23 @@
 const Colors = {
+  // Green
+  lightestGreen: '#C9E3D9',
+  lighterGreen: '#48AC69',
   primaryGreen: '#008550',
   darkerGreen: '#004E14',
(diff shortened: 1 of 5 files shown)


assets/Colors.js
```diff
@@ -1,10 +1,23 @@
...
+  // Text
+  activeText: 'rgba(0, 0, 0, 0.8)',
+  secondaryText: '#999999',
+  // Other
   base: '#BDBDBD',
-  black: '#000000',
-  white: '#FFFFFF'
+  light: '#DEDEDE',
+  lighter: '#EBEBEB',
+  lightest: '#FFFFFF',
```
aivantg: why is this not just named "white" 🙄

PR #31, opened 2020-02-27
Resources Screen

kennethlien:

# What's new in this PR

Added the community resources screen! Click on the hamburger menu to navigate to it. It's still lacking styling and needs to implement base components, so that's work for next time.

UPDATE: It now seperates links into categories, added card component for resources.

### Related PRs

This page is super simple, but most the stuff was built in the News screen PR.

## How to review

Make sure the links work, and suggest styling changes.

## Next steps

Add style, use reusable components!

UPDATE: STYLE

Clicking went well! Scrolling still works!

CC: @tommypoa 
@wangannie

### DIFF
--- a/screens/resources/ResourceScreen.js
+++ b/screens/resources/ResourceScreen.js
@@ -1,27 +1,47 @@
 import React from 'react';
-import { TouchableOpacity, Button, View } from 'react-native';
-import { Linking } from 'react-native';
 import { TopText } from '../../styled/resources';
+import { TouchableOpacity, ScrollView, View, Button } from 'react-native';
+import getResources from '../../lib/resourceUtils';
+import { Title } from '../../components/BaseComponents';
+import ResourceCard from '../../components/resources/ResourceCard';
 
-const resources = [
-  {
-    name: 'DC Central Kitchen',
-    url: 'https://dccentralkitchen.org/'
-  },
-  {
-    name: 'Healthy Corners Initiative',
-    url: 'https://dccentralkitchen.org/healthy-corners/'
-  },
-  {
-    name: 'Important Link',
-    url: 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'
-  }
-];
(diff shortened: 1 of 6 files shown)


screens/resources/ResourceScreen.js
```diff
@@ -0,0 +1,70 @@
...
+class ResourceScreen extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  render() {
+    return (
+      <View>
+        <TouchableOpacity
```
anniero98: You probably want to put this in at least a styles object, so that it's easily exported to a stylesheet and then can be refactored to `styled-components`. Since the Resources screen is going to be more complex than we thought, I think it's worth it to pull this out and add it to the `styled/` folder - you can make it `resources.js` (-:

styled/resources.js
```diff
@@ -0,0 +1,40 @@
+import styled from 'styled-components/native';
+import { PoppinsText } from './shared';
```
anniero98: This is deprecated - import from `BaseComponents` if you want to extend for text

styled/resources.js
```diff
@@ -0,0 +1,40 @@
...
+export const DateText = styled(PoppinsText)`
+  font-weight: bold;
+`;
+
+export const ContentContainer = styled.View`
+  flex-direction: column;
+  flex: 4;
+`;
+
+export const ContentText = styled(PoppinsText)`
```
anniero98: Let's get rid of all of these besides the `TopText`, since we aren't using any of those. Try and keep only what's being used right now !

screens/resources/ResourceScreen.js
tommypoa: Given this week's Figma and our latest idea of the Resources page, I think it would be a good idea to create a new component - possibly something like `ResourceCard` - in addition to the `styled-components` in `resources.js`! The main reason is that all the resources links are in the format (style + content-wise).

PR #42, opened 2020-03-19
Navigation Button restyling + Misc. style fixes

kennethlien:

# What's new in this PR

Added:
- New base component: NavButton
- Changed screens to match Figma buttons
- Changed the header of resources slightly, going to ask Ace some followup before making it match Figma

NOTE: If you want to test this, first find all instances of "zindex" and replace it with "zIndex", since prettier changed everything when I committed.

### Online sources

### Related PRs
This PR is dependent on the Hamburger menu PR, since it uses a lot of things from there and also is branched off it.

## How to review
BaseComponents -> styled/rewards.js, and then anywhere

## Next steps

Tell prettier to stop changing zIndex to zindex!!!

### Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-03-18 at 20 57 29
Simulator Screen Shot - iPhone 11 Pro - 2020-03-18 at 20 57 36
Simulator Screen Shot - iPhone 11 Pro - 2020-03-18 at 20 57 07

CC: @12aschen  @wangannie


styled/rewards.js
```diff
@@ -1,13 +1,18 @@
...
-import { Subhead, Caption } from '../components/BaseComponents';
+import { Subhead, Caption, NavButton } from '../components/BaseComponents';
 import Colors from '../assets/Colors';
 
 export const Container = styled.View`
   flex: 1;
   background-color: #fff;
 `;
 
+export const BackButton = styled(NavButton)`
```
wangannie: It took me a while to realize why this had a green background. I feel like this could get messy setting background colors when they mostly don't have backgrounds...see my comment in BaseComponents about this

components/BaseComponents.js
```diff
@@ -12,6 +12,21 @@ export const TextButton = styled.Text`
   color: ${props => props.color || Colors.primaryGreen};
 `;
 
+export const NavButton = styled.TouchableOpacity`
+  background-color: #fff;
```
wangannie: Your thinking was really good to make `NavButton` a Base Component considering how often it's reused! Considering that this button has no background in most use cases (everywhere except hamburger), you should make it take in background color as a prop and default to transparent. You can maybe do this by using an RGBA value `rgba(255, 255, 255, 0)`.

styled/hamburger.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components/native';
+import { NavButton } from '../components/BaseComponents';
+import { TouchableOpacity } from 'react-native';
+
+//TODO: Fix top to match figma once we figure out iPhone alignment
+export const HamburgerButton = styled(NavButton)`
+  top: 63px;
+  left: 16px;
```
wangannie: I think you can solve this by moving `<Hamburger navigation={this.props.navigation} />` into the `MapView`, right above `SearchBar`. Then you can make the `top` here the same as the top margin on `SearchBar` (20px I believe), and it should stay lined up with the height of the search bar, which has mostly been looking fine on different kinds of screens.

styled/rewards.js
wangannie: ALSO then you wouldn't have to extend this! You'd just use props when you need a button with a background.

components/Hamburger.js
```diff
@@ -1,38 +1,18 @@
...
-import {
-    TouchableOpacity,
-    Button
-} from 'react-native';
-
+import { TouchableOpacity, Button } from 'react-native';
+import { FontAwesome5 } from '@expo/vector-icons';
+import { NavButton } from '../components/BaseComponents';
+import { HamburgerButton } from '../styled/hamburger';
+//TODO: ADD FONTAWESOME ICON
```
wangannie: ??

styled/rewards.js
```diff
@@ -1,13 +1,18 @@
...
+  background-color: ${Colors.primaryGreen};
+  border-color: ${Colors.primaryGreen};
+`;
```
wangannie: It took me a while to realize why this had a green background. I feel like this could get messy setting background colors when they mostly don't have backgrounds...see my comment in BaseComponents about this 

ALSO then you wouldn't have to extend this! You'd just use props when you need a button with a background.

PR #45, opened 2020-03-24
Navigation styling

wangannie:

# What's new in this PR
* Created `NavHeaderContainer` and `NavTitle` base components to standardize navigation headings and reduce repeated code. All text and buttons should be consistently positioned!
* Removed default navigation bars and added custom headings (with drop shadows) and new NavButtons from #42 on:
  * Welcome
  * Login
  * Signup
  * Store List (also added new CANCEL button consistent with Figma)
  * Products List (`ProductsScreen`)
  * Resources
  * Map Screen (transparent container)
* Cleaned up some deprecated text styles and styled-components
* Fixed a scrolling bug on the products list

### Related PRs
Inspired by the standardized NavButton base component from #42 !

## How to review
Make sure everything looks good on different screen sizes/aspect ratios.

## Next steps
_Resolved in 3/26 Update:_

- [x] I want to standardize the heading styling somehow (maybe as a base component). I'm leaving this as a draft because I need to go back and clean this up and see how I can reduce the repeated code.
- [x] I also need to fix the Rewards heading bar, but that involves fixing the tab styling.

- [x] Figure out SafeAreaView or something with `react-native-device-info` to adjust the top bar margin depending on if the device has a notch. It looks really uncomfortably big on iPhone 8 simulator right now. **Update: resolved this in #56** 

* Tested on a few different screen size simulators.

### Screenshots
* Styled back buttons on login and signup:
image

* Store list styled heading with new cancel button
image

* Updated Resources heading title bar
image

* No more white bar here! Map is now full screen.
image


screens/map/MapScreen.js
```diff
@@ -261,7 +260,7 @@ export default class MapScreen extends React.Component {
             <Subhead color={'#fff'}> Your Rewards </Subhead>
           </View>
         </TouchableOpacity>
-      </SafeAreaView>
+      </View>
```
kennethlien: This looks so much better :D

screens/resources/ResourcesScreen.js
```diff
@@ -45,13 +48,12 @@ class ResourcesScreen extends React.Component {
...
-
-        <View style={{ height: 106 }}>
-          <TopText>Resources</TopText>
-        </View>
+        <NavHeaderContainer>
+          <NavButton onPress={() => this.props.navigation.goBack(null)}>
+            <FontAwesome5 name="arrow-left" solid size={24} />
+          </NavButton>
+          <NavTitle>Resources</NavTitle>
+        </NavHeaderContainer>
```
tommypoa: Looks fine on iPhone 11 but on iPhone 8 it the header/"Resources" isn't center-aligned. There is also a gap at the top of the header (also in Points History) that presents itself in < iPhone X versions - not sure if this is a problem/deviation from Figma but just flagging.
image

screens/rewards/RewardsScreen.js
```diff
@@ -144,9 +144,9 @@ export default class RewardsScreen extends React.Component {
     return (
       <Container>
         <TopTab>
-          <BackButton onPress={() => this.props.navigation.goBack()}>
+          <NavButton onPress={() => this.props.navigation.goBack()}>
             <FontAwesome5 name="arrow-down" solid size={24} color="white" />
-          </BackButton>
+          </NavButton>
```
tommypoa: Button isn't center-aligned / seems to be flushed to the bottom of the header, and looks like it is sticking right above the words "Healthy Rewards".
image

screens/auth/LoginScreen.js
```diff
@@ -146,6 +148,9 @@ export default class Login extends React.Component {
   render() {
     return (
       <AuthScreenContainer>
+        <BackButton onPress={() => this.props.navigation.goBack(null)}>
+          <FontAwesome5 name="arrow-left" solid size={24} />
+        </BackButton>
```
tommypoa: Is this BackButton supposed to be positioned somewhere at the top left of the screen? Fine in iPhone 11 but on iPhone 8 it appears to be closer to "Log in" than the top.
image

screens/resources/ResourcesScreen.js
wangannie: Good catch! Both of these issues were later fixed in #56

screens/rewards/RewardsScreen.js
wangannie: Interesting! It is actually properly aligned but does not appear to be because the icon has a 40px width and some padding, making the icon itself only 22px centered in the 40px so it appears misaligned. But thank you for noticing this! I will adjust it in #56

screens/auth/LoginScreen.js
wangannie: This is also adjusted to be higher in #56 for the keyboard covering issues. 👍

PR #47, opened 2020-03-26
Link store address to directions in maps app

wangannie:

# What's new in this PR
* Tap the address line of a store to either:
  * Get directions in an external maps app
  * Copy the address to clipboard

### Online sources
n/a
### Related PRs
n/a
## How to review

* Make sure to set your simulator location to be near the DC area (otherwise it might open similarly-named stores...not sure how to fix this)
image

* Try opening directions for a bunch of different stores—it searches for the nearest store by address closet to the store's longitude, there may be some exceptions where it doesn't find the right store. I ended up abandoning longitude/latitude bc it was really jank for Android, so I went with a query string that may be hit or miss.
## Next steps

* Look into [Action Sheets](https://github.com/expo/react-native-action-sheet) instead of Alerts 

* Tried to simulate cross-platform compatibility by setting it to Google maps for Android 
### Screenshots
https://www.loom.com/share/836be07d58664c77a39bdd62dfdc05de
image
image


components/store/StoreCard.js
```diff
@@ -51,14 +56,41 @@ function StoreCard({ store, callBack, seeProduct }) {
...
+      <TouchableOpacity
+        onPress={() =>
+          Alert.alert(
+            name,
+            address,
+            [
+              {
+                text: 'Get Directions',
+                onPress: () =>
+                  openMap({
```
tommypoa: I tested this on my phone (iOS) but because I had my default Maps app deleted, it prompted me to download the default app and is not able to have it open on other map apps (GoogleMaps, Waze etc.). Am not sure if this will be a big problem (I might be the rare few who deletes the default Map app) but just to note the edge case! 

As a reference, the `react-native-open-map` package (https://www.npmjs.com/package/react-native-open-map) might give some additional flexibility if we want to account for this case.

wangannie: WOW thank you `react-native-map-link`(what that's based off of) ended up being exactly what I needed! Wish I had found this earlier.

wangannie:
Updates:
* Switched to a different package `react-native-map-link` for better cross-platform compatibility (and to accommodate those who uninstall their system default map apps).
image
* Moved copy to clipboard to work on a long press instead  of a click (didn't want too many nested system dialogs)
image
* Cleaned everything up into helper functions based on @tommypoa 's  suggestion

components/store/StoreCard.js
```diff
@@ -19,8 +20,24 @@ import StoreProductButton from './StoreProductButton';
...
  * */
 
-function StoreCard({ store, callBack, seeProduct }) {
-  const { name, hours, address, distance, ebt, rewards } = store;
+export default function StoreCard({ store, callBack, seeProduct }) {
+  const { name, hours, address, distance, ebt, rewards, lat, long } = store;
+
+  const writeAddressToClipboard = () => {
+    Clipboard.setString(address);
+    Alert.alert('Copied to Clipboard!', address);
```
kennethlien: Could this be a popup that just fades away? It seems less obtrusive.

PR #49, opened 2020-03-26
Truncate long store names

tommypoa:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR

[//]: # "Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."
- Store names that are too long for the width of the user's device will be truncated. 
## Relevant Links
- NIL

## How to review

[//]: # "The order in which to review files and what to expect when testing locally"
- Run through code to check for code quality
- Give suggestions on better way to implement / improvements to existing implementation
## Next steps

[//]: # "What's NOT in this PR, doesn't work yet, and/or still needs to be done"
- Possibly making css fix - currently blocked because of incompatibility with React Native Styled Components
## Tests Performed, Edge Cases

[//]: # "Hopefully we will add a testing suite/CI soon, but until then note down the steps you took to test locally"
- NIL
### Screenshots

[//]: # "Add screenshots of expected behavior - GIFs if you're feeling fancy!"
image

CC: @wangannie @kennethlien 

[//]: # "This tags in both Annies as a default. Feel free to change, or add on anyone who you should be in on the conversation."

PR #51, opened 2020-03-27
[Updated] Truncate long store names

wangannie:

# What's new in this PR
* Long store names are trimmed with ellipses, adjusting depending on:
  * if there is an EBT tag
  * if it is displayed with a 'See Products' button.
* This is taking the same idea from #49 but adjusted to factor in all considerations, not just limiting it to the minimum possible width (from the old calculation that assumed there was always the 'See Products' button and EBT tag). This system should be more consistent overall.

### Related PRs
Originally branched off of (but now replacing)  #49 

## How to review
* Look at store names of various lengths and test with different simulator screen sizes.

## Next steps
* Consider modifying the design to have a smaller font size title and narrower button so less information has to be truncated. 

### Screenshots
* Titles are adjusted according to the surrounding components. Minnesota Food Mart (with no EBT tag) is able to take more width than those with EBT tabs.
image

* In map screen, particularly long titles will shrink to display the full line. This is only compatible with iOS. 
image

* Titles still take up the full width if there isn't a See Products button
image

CC: @anniero98 @tommypoa


lib/mapUtils.js
```diff
@@ -66,3 +66,12 @@ export const getProductData = function async(store) {
...
   });
 };
+
+// Necessary to take into account different screen sizes
+// 32px: Width of left and right margins
+// 12px: Extra space for padding between items
+// 130px: Size of 'See Products' button
+// 48px: Size of EBTBar
+export const getMaxWidth = function sync(screenWidth, ebt, seeProducts) {
+  return screenWidth - 32 - 12 + (ebt ? -48 : 0) + (seeProducts ? -130 : 0);
```
tommypoa: We should add a @todo to see how we can minimize hard-coded values here, and throughout the app (if any exists now)! No good solution as of yet - might just require specific/clear documentation during handoff.

PR #48, opened 2020-03-26
Airtable refactor, bug fixes, Auth overhaul

anniero98:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR

Refactoring our Airtable code in this repo to use `airtable-schema-generator` (now that it's been updated to 1.3.3 with some nice features!).

### REQUIRES `.env.development`/`production` to be changed from `AIRTABLE_API_KEY` to `REACT_APP_AIRTABLE_API_KEY`
...for compatibility for v1.3.3

Has the same Airtable schme updates as https://github.com/calblueprint/dccentralkitchen-clerks/pull/29; Points History screen will not load without transaction columns being changed to `Total Sale`.

- Removed code in `common.js` since `BASE` will no longer be used

### Map
- Bug: initial store showed the incorrect products
- Edge case: If no products are linked to a store, the key `productIds` will not exist on the `store` object, so special handling is required

### Navigation
- Cleanup: refactored `AppNavigator` by extracting `DrawerContent` as a class, and moving `AuthStack` to `StackNavigators`

### News
- Bug: `NewsScreen` displayed `Created` date, not `Post Date`

### Rewards
- Deprecated: removed pull-to-refresh implementation to declutter the file. We can add it back from a different trigger if necessary. Removed deprecated props which were not being used in `RewardsHome` and `PointsHistory` components.
- Cleanup: Made `Overline` components normally-cased (since the styled component uses `text-transform`)
- Cleanup: Added `isLoading` to `DrawerContent` and `RewardsScreen`. This is probably the most crucial place to add a loading screen (@ace) imo; it's noticeably laggy here.
- Cleanup: Add `Rewards.js` to `constants` so things aren't hardcoded.

### Auth
Integrated @tommypoa 's changes from `material-ui` branch

### SignUp
- Fixed SignUp's duplicate customer check to actually work, but no design exists for displaying the error right now (@12aschen)
- Replaced the `name, nameError` etc in `this.state` to use an object similar to `indicators`.
- Changed behavior of `handleErrorState` to update error messages from `validate.js` instead of being hardcoded into `AuthField`
- Added an error message and error state for `name`
- Made `onTextChange` hooks async so that error handling would happen _after_ values updated in state; and handled permissions
* Note that `onBlur` is still required for the edge case of user tapping in then tapping out
- Added margin to the SignUp button

### Login
- Added error message and no error message but error indicators for that state.
[...]

PR #72, opened 2020-04-05
Default store, recenter location button

tommypoa:

# What's new in this PR
- Dollar Plus Howard will be set as the default store if
   - User is more than 100 miles away
   - Location permission is disabled
- Design changes for the above default case implemented
- Recenter location button added
   - Implemented in #11 
- Store list re-orders if you change your location, then hit the "center location" button

## Relevant Links

### Related PRs

#11 
## How to review

- Manually set location to be far away and see if the default case is triggered/look for edge cases
- Visual alignment to Figma
- Test recenter button

## Next steps

- Ordering of storelist during default option (distance from Dollar Plus Howard/default store)
- Position of recenter button

### Screenshots

- Recenter button right below hamburger for now. Icon of recenter taken from FontAwesome.
image

CC: @wangannie


screens/map/MapScreen.js
```diff
@@ -204,6 +225,7 @@ export default class MapScreen extends React.Component {
             </Subhead>
           </SearchBar>
         </NavHeaderContainer>
+        <CenterLocation callBack={() => this._findCurrentLocationAsync()} />
```
wangannie, line 228: Definitely nonessential but are there easy ways to make this transition smoother?

screens/auth/SignUpScreen.js
```diff
@@ -9,19 +9,19 @@ import AuthTextField from '../../components/AuthTextField';
 import {
   BigTitle,
   ButtonLabel,
-  FilledButtonContainer
+  FilledButtonContainer,
```
anniero98, line 12: Not sure why these changes didn't happen in #48 but it's nbd I think

screens/map/MapScreen.js
```diff
@@ -79,7 +86,13 @@ export default class MapScreen extends React.Component {
         longitude: location.coords.longitude,
         ...deltas,
       };
-      this.setState({ locationErrorMsg: null, location, region });
+      // Don't re-animate if we're using the default store
```
anniero98, line 89: I tried to make this work with the default store, but the resetting of the location in `orderStoresByDistance` for only the default store produced some really odd behavior, so I think it's ok to lose the animation for the default store.

PR #85, opened 2020-04-09
Justin/sentry second try

JustinMi:
# README:

There was a whole lotta merge conflicts in my previous PR and I think remaking the branch is faster than trying to resolve all of them, so here's a new PR haha. Everything below this message is straight copy-pasted from the previous PR's description. The diffs are cleaner in this PR too. 

# What's new in this PR

 I integrated Sentry 🤖

In `App.js`, I added a couple of lines at the top to initiate Sentry. In `app.json`, I added some specifications for a post-publish hook. In `SignUpScreen.js` and `LoginScreen.js`, I added a snippet to capture the current user's ID. I also added the Sentry config info to `.env.example`.

### Online sources

https://forum.sentry.io/t/set-release-after-init/6759/2

## How to review

To test, you gotta first install Sentry. I think you can just `yarn install` on my branch, but if that doesn't work, then do `yarn add sentry-expo`. 

Next, log into Sentry using the login info pinned in #chaos. Navigate to the `dccentralkitchen` project, where all the issues are logged. That's the page where you will check if Sentry actually catches errors.

Next, add Sentry config info into to your `.env.development`. You can find them in `.env.example`. They look like this:
```
...
SENTRY_ORG='...'
SENTRY_PROJECT='...'
SENTRY_AUTH_TOKEN='...'
```
You can use the `SENTRY_AUTH_TOKEN` is an org-wide token on Sentry, not user-specific. That way if we add/remove people to Sentry we don't gotta fiddle with token stuff. To find it: assuming you have access to the Blueprint org on Sentry, then go to sentry.io, and hit: `Settings` -> `Developer Settings` -> `Internal Integrations` -> `DC Central Kitchen Auth Token` -> `Tokens` to find it.

Finally, create an error. One thing I did was drop `<button onClick={methodDoesNotExist}>Break the app</button>;` somewhere in `WelcomeScreen.js`, and then I opened the app the in Expo simulator. You should see the app display a red error screen, and the error will also pop up as an Issue on the Sentry page. Note that if you repeatedly trigger the same error it logs under the same Issue, it doesn't create a new Issue. That confused me at first b/c it's not immediately obvious in the UI that the Issue updated.

## Next steps

I'll need to add this to `dccentralkitchen-clerks` as well

### Screenshots

What I did to elicit an error:
image

What the iOS simulator should show: 
image

What the terminal should show:
image

What an Issue looks like in Sentry: 
image

CC: @anniero98 @wangannie


JustinMi:
@FranCogMents I added the code snippets capturing the user 👍

 You mentioned something about using `Sentry.captureException` as well--is that necessary? I may be wrong but I thought Sentry automatically captured all errors, so manually capturing exceptions is not necessary.

screens/auth/SignUpScreen.js
```diff
@@ -158,6 +159,13 @@ export default class SignUpScreen extends React.Component {
...
       await this._asyncSignUp(customerId);
+      // register this user in the Sentry logger
+      Sentry.configureScope((scope) => {
+        scope.setUser({
+          id: customerId,
+          phoneNumber: formattedPhoneNumber,
+        });
+      });
     } catch (err) {
       console.error('[SignUpScreen] (handleSubmit) Airtable:', err);
```
FranCogMents, line 170: sentry sdk automatically captures unhandled promises and exceptions -- but we are want to capture and send the errors that are already handled/caught by the app, which is why we have to explicitly call `Sentry.captureException`

Wilson's PR for reference for javascript examples: https://github.com/calblueprint/unloop/pull/73/files

We should be logging when:
1. sections of the codebase are catching errors through `.catch(error)` (i.e airtable queries)
2. sections of the app where there is an error state being logged
- perfect candidate of #2 is the `customers.length >1` case in the `LoginScreen` where it's a database error that isn't an uncaught promise or exception but we need to send

Example of adding extra information for context in capturing the exception
```
Sentry.configureScope(function(scope) {
          scope.setExtra('screen', 'SignUpScreen');
          scope.setExtra('action', 'handleSubmit');
        });
        Sentry.captureException(err);
```

screens/auth/LogInScreen.js
```diff
@@ -92,6 +94,13 @@ export default class LogInScreen extends React.Component {
         await updateCustomerPushTokens(customer, token);
         // Log in
         await this._asyncLogIn(customer.id);
+        // register this user in the Sentry logger
+        Sentry.configureScope(scope => {
```
wangannie, line 98: Need to move this before line 95. Right now it doesn't seem to be logging users. Same in Sign Up

App.js
```diff
@@ -5,8 +5,30 @@ import * as Font from 'expo-font';
...
+// from https://forum.sentry.io/t/set-release-after-init/6759/2
+// may be able to turn this code block off in production
+setTimeout(async () => {
+  Sentry.configureScope((scope) => {
+    scope.addEventProcessor((event) => {
+      event.release = 'newRelease';
+      return event;
+    });
+  });
+  await Sentry.captureMessage('Something Broke');
```
wangannie, line 26: Do we still need this? It seems to run every time regardless of if there's an error, and doesn't seem to give any user info

screens/auth/LogInScreen.js
wangannie: Also even with it moved, it doesn't seem to be picking up the phone numbers?

App.js
```diff
@@ -5,8 +5,30 @@ import * as Font from 'expo-font';
 import React, { useState } from 'react';
 import { Platform, StatusBar, StyleSheet, View } from 'react-native';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
+import * as Sentry from 'sentry-expo';
 import AppNavigator from './navigation/AppNavigator';
 
+Sentry.init({
+  dsn: 'https://dacd32167a384e189eab16e9588c0e67@sentry.io/5172575',
+  enableInExpoDevelopment: true,
+  debug: true,
```
wangannie, line 14: For production do I just turn both of these to false?

screens/auth/LogInScreen.js
FranCogMents: I verified that this does register a user and associates any future sentry errors (although I had to hardcode it to verify) but it seems that `getAllCustomers` kept failing with the test accounts (both on prod + test), which is why it currently was not registering any information.
```
Sentry.configureScope(scope => {
         scope.setUser({
          id: '55',
           phoneNumber: formattedPhoneNumber,
       });
      }); 
```

wangannie:
FYI Adding this 
`        <FilledButtonContainer
          style={{ marginTop: 104 }}
          width="100%"
          onPress={() => this.doesntwork()}>
          <ButtonLabel color={Colors.lightest}>break</ButtonLabel>
        </FilledButtonContainer>`

to `RewardsHome.js` helped with testing to see if Sentry captured the user info with the error.

screens/auth/LogInScreen.js
```diff
@@ -88,6 +78,13 @@ export default class LogInScreen extends React.Component {
       // Returns empty array if no customer is found
       if (customers.length === 1) {
         const customer = customers[0];
+        // register this user in the Sentry logger
+        Sentry.configureScope(scope => {
+          scope.setUser({
+            id: customer.id,
```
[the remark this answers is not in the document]
anniero98: I would guess this is a privacy/security concern

screens/auth/SignUpScreen.js
```diff
@@ -162,6 +171,7 @@ export default class SignUpScreen extends React.Component {
       // Otherwise, add customer to Airtable
       const customerId = await this.addCustomer();
       this._clearState();
+      // register this user in the Sentry logger
```
anniero98, line 174: nit: remove

screens/auth/LogInScreen.js
```diff
@@ -48,6 +50,7 @@ export default class LogInScreen extends React.Component {
   // to be the user ID and then navigates to homescreen.
   _asyncLogIn = async userId => {
     await AsyncStorage.setItem('userId', userId);
+    Sentry.captureMessage('Log In Successful');
```
anniero98, line 53: If it's possible to log `userId` in this message, it'll be a much more valuable message; also I might guess that `Log In Unsuccessful` for a particular case would be rarer (hopefully) and something we are actually looking for

App.js
JustinMi: I think `enableInExpoDevelopment` is specifically for Expo to be able to interact with Sentry.. probably not necessary in production. `debug` can be turned off in production.

screens/auth/LogInScreen.js
wangannie: It still does capture the user info in line 95 so these messages are also attached to the user details.

screens/auth/LogInScreen.js
FranCogMents:
```
 Sentry.configureScope(scope => {
          scope.setUser({
           id: `${customerId}`,
            phoneNumber: formattedPhoneNumber,
        });
       }); 
```

@wangannie  @JustinMi -- let's move this as an else case to where it is guaranteed that it is a succcesful log-in since we will only need to set the user once: https://github.com/calblueprint/dccentralkitchen/blob/8c713862a78b5dbf3ceef6d5b951c5c2fa8113b1/screens/auth/LogInScreen.js#L124

Because the customerId is not saved in state, you would need to make customerId + customer `var variables` so they can be accessed in this else case

```
if (error !== '') {
        logAuthErrorToSentry({
          screen: 'LogInScreen',
          action: 'handleSubmit',
          attemptedPhone: formattedPhoneNumber,
          attemptedPass: password,
          error,
        });
   } else {
     Sentry.configureScope(scope => {
          scope.setUser({
           id: `${customerId}`,
            phoneNumber: formattedPhoneNumber,
        });
       }); 
}
```

screens/auth/SignUpScreen.js
```diff
@@ -129,7 +131,14 @@ export default class SignUpScreen extends React.Component {
         points: 0,
         pushTokenIds: pushTokenId ? [pushTokenId] : null,
       });
-
+      Sentry.configureScope(scope => {
+        scope.setExtra('setUserScreen', 'SignUpScreen');
```
FranCogMents, line 135: let's remove this -- this will be set and reported when an error is caught.

When we call `configureScope`, any changes we add will be kept on successive calls to `configureScope` so, for example if we kept the call here `scope.setExtra` and there was an error in a different screen, our Sentry log would report this extra info "setUserScreen: signUpScreen" when an error actually occured in rewards.

so let's remove this line, I'll add comments on changes to the logUtils.js function calls

navigation/DrawerContent.js
```diff
@@ -24,6 +25,14 @@ class DrawerContent extends React.Component {
       } else {
         customer = { name: 'Guest' };
       }
+      Sentry.configureScope(scope => {
+        scope.setExtra('setUserScreen', 'DrawerContent');
```
FranCogMents, line 29: @JustinMi @wangannie Let's remove the `scope.setExtra` here as well, this should only be added when reporting a message or exception using `withScope` which will attach the `setExtra` information only within the the next call to .captureException or .captureMessage

scopes have been a journey for me to learn, here is what I'm referencing: [scopes](https://docs.sentry.io/enriching-error-data/scopes/?platform=javascript#configuring-the-scope)

lib/logUtils.js
```diff
@@ -0,0 +1,21 @@
...
+  Sentry.withScope(scope => {
+    scope.setExtra('screen', screen);
+    scope.setExtra('action', action);
+    Sentry.captureException(error);
+  });
+}
+
+export function logAuthErrorToSentry(errorObject) {
+  const { screen, action, attemptedPhone, attemptedPass, error } = errorObject;
+  Sentry.withScope(scope => {
```
FranCogMents, line 14: NICEEEE use of `withScope`! lovvinnn it!

screens/auth/LogInScreen.js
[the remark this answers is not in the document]
FranCogMents: @wangannie @anniero98 
For any sensitive data -- sentry will automatically scrub the data and those options can be modified on the dashboard -- which affects how it appears in logs -- passwords for example will automatically be redacted. username should be fine -- I think i would be more inclined to to not log the phone number if it's not necessary, since that's information that is personally identifiable outside of the context of the application, vs the id and username is DC Central app generated content that identifies a user, which can still be used to effectively debug.

 if it's not necessary to log the phone number, you can if you can just log the `id` + `username` both should be enough to effectively debug but those are my initial concerns

screens/auth/LogInScreen.js
FranCogMents: @wangannie @JustinMi -- with Firebase analytics tracking these changes, i would also suggest removing this and having Sentry focus more on capturing error states

App.js
```diff
@@ -5,8 +5,17 @@ import * as Font from 'expo-font';
...
 import { SafeAreaProvider } from 'react-native-safe-area-context';
+import * as Sentry from 'sentry-expo';
 import AppNavigator from './navigation/AppNavigator';
 
+Sentry.init({
+  dsn: 'https://dacd32167a384e189eab16e9588c0e67@sentry.io/5172575',
+  enableInExpoDevelopment: true,
+  release: 'v1.1.0',
+  debug: true,
+  environment: 'development',
```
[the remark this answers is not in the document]
JustinMi: Ok, I'll make a note to add this in the next PR

screens/auth/SignUpScreen.js
```diff
@@ -129,7 +131,14 @@ export default class SignUpScreen extends React.Component {
...
+        scope.setUser({
+          id: customerId,
+          username: name,
```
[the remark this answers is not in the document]
JustinMi: ok I'll remove the username for this PR, we can decide in future PRs

screens/auth/SignUpScreen.js
```diff
@@ -163,9 +182,17 @@ export default class SignUpScreen extends React.Component {
...
       this._clearState();
+      // register this user in the Sentry logger
       await this._asyncSignUp(customerId);
     } catch (err) {
       console.error('[SignUpScreen] (handleSubmit) Airtable:', err);
+      logAuthErrorToSentry({
+        screen: 'SignUpScreen',
+        action: 'handleSubmit',
+        attemptedPhone: null,
+        attemptedPass: null,
```
FranCogMents, line 193: making a note on next round of Sentry PR's to make/modify logging for SignUp

PR #93, opened 2020-04-16
Additional reward cards for point history

tommypoa:

# What's new in this PR

- Implemented 2 new types of reward cards for the Points History page
   - Reward redeemed card
   - Reward unlocked card 
## Relevant Links
- NA

### Related PRs

- NA
## How to review

- See if it looks like the designs in Figma
- Suggest ordering for the 3 types of reward cards

## Next steps

- Perhaps adding monthly date filtering

### Screenshots

image

CC: @anniero98 @wangannie


components/rewards/Transaction.js
```diff
@@ -11,30 +12,77 @@ import CircleIcon from '../CircleIcon';
...
+          iconColor={Colors.primaryGreen}
+          circleColor={Colors.lightestGreen}
+        />
+        <ContentContainer>
+          <Caption color={Colors.secondaryText}>
+            {`${date.toLocaleDateString('en-US', options)} • ${storeName}`}
+          </Caption>
+          <Subhead>{`${pointsEarned} points earned`}</Subhead>
+          <Caption color={Colors.secondaryText}>
+            {`for ${displayDollarValue(totalSale || 0)} of healthy products`}
```
wangannie, line 44: I know it was like this before but would this make more sense to be subtotal?

wangannie: Oh nevermind I guess it makes sense that the number of points directly corresponds to the total sale (bc the rewards points are deducted?) but this language might not fit since they did actually pay more for the total healthy products...if that makes sense
cc @12aschen ?

screens/auth/AuthLoadingScreen.js
```diff
@@ -21,6 +21,7 @@ export default class AuthLoadingScreen extends React.Component {
 
     // Auth/App testing purpose
     // this.props.navigation.navigate('Auth');
+    // await AsyncStorage.setItem('userId', 'recTSlhgMXVkpoRyv');
```
wangannie, line 24: Is this for bypass? Don't forget to take it out

navigation/AppNavigator.js
```diff
@@ -31,11 +32,11 @@ function DrawerNavigator() {
         component={StoresStackNavigator}
         options={{ title: 'Stores' }}
       />
-      {/* <Drawer.Screen
+      <Drawer.Screen
         name="Rewards"
         options={{ title: 'Points History', drawerLockMode: 'locked-closed' }}>
         {props => <RewardsScreen {...props} tab={1} />}
-      </Drawer.Screen> */}
+      </Drawer.Screen>
```
wangannie, line 39: This was supposed to be left out for Guest Mode since there are no tabs in Reward for Guest Mode. We should take it out again.

components/rewards/Transaction.js
```diff
@@ -11,30 +12,77 @@ import CircleIcon from '../CircleIcon';
...
+          <ContentContainer>
+            <Caption color={Colors.secondaryText}>
+              {`${date.toLocaleDateString('en-US', options)} • ${storeName}`}
+            </Caption>
+            <Subhead>{`$5 reward redeemed`}</Subhead>
+          </ContentContainer>
+        </TransactionCard>
+      ))}
+
+      {/* Display rewards unlocked */}
```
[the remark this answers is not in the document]
wangannie: So the order (top to bottom, most to least recent) would be Unlocked, Earned, Redeemed

components/rewards/Transaction.js
anniero98: It's probably easiest to match what we do in Clerk (which I cannot currently recall) but I would guess this is a quacetion that we can defer. Just don't forget to check in on it 😅

PR #96, opened 2020-04-17
Added store hours and store details

thumn:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR
- Store cards now display whether the store is currently open or what time it will next be open
- The store card now navigates to a separate store card on click that provides information on the store's address, phone number, hours by day of the week, and accepted programs (NOTE: accepted programs is not finished)
- Detailed store hour schedules bold the current day and hours

[//]: # "Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

## Relevant Links
N/A

### Online sources
[Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)

[//]: # "Optional - copy links to any tutorial or documentation that was useful to you when working on this PR"

### Related PRs
N/A

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

## How to review
Navigate through stores and make sure each store's card can successfully navigate to its corresponding store screen. The current day should be correctly bolded, and the store open status summary should correctly reflect whether the store is open or when it'll next be open. For example, Yang Market has a particular complicated store hours input.

[//]: # "The order in which to review files and what to expect when testing locally"

## Next steps
Start work on tags for accepted programs in the stores list. This will also be applied to the store screens.

NOTE: I'm getting a "each child in a list should have a unique 'key' prop" error -- does anyone know how to fix this? The file is `StoreHours.js`. I also can't figure out how to make the `<StoreDetailText>` primary green on the `StoreCard`, but it works fine on the store screen. (Might be passing in the wrong prop, but the component seems to take in `color` as a prop so I'm not too sure.)

Also, I sometimes get a weird "store is undefined" error on `MapScreen` that goes away if I reload the app. Sadly this might end up circling back to the `useFocusEffect` hook issue/refactoring task we set aside :(

[//]: # "What's NOT in this PR, doesn't work yet, and/or still needs to be done"

## Tests Performed, Edge Cases
Tested with hard-coded store hour inputs:
`9am-8pm Tu-Sat, 9am-7pm Sun, Closed Mon`
`8am-7pm Daily`
`Open 24/7`
`7am-Midnight Daily`

[...]

PR #98, opened 2020-04-18
Added accepted programs tags

thumn:

# What's new in this PR
`StoreCard` and `StoreDetailsScreen` now display:
- Ward number
- Distance from your current location
- Tags for accepted programs

### Online sources
[`react-native-paper` Chip](https://callstack.github.io/react-native-paper/chip.html)

### Related PRs
#96 

## How to review
Navigate to different store cards through the map, in the `StoreList` screen, and through the search feature. Make sure tags are showing up correctly in the `StoreCard` and that everything looks visually centered in the `StoreDetailsScreen`.

For example, it looks like Grubbs is the only store that has the WIC tag. Nam's is currently the only store that has the Healthy Rewards tag.

## Next steps
I could not for the life of me figure out how to stop the tag descriptions from overflowing the page on the `StoreDetailsScreen` even with the use of `flexWrap: wrap`. I'm submitting this PR anyway because we have bug squashing tomorrow, and all that's left is really a bunch of visual bugs. The main functionality of the tags is finished, so if we need that by tomorrow, here it is!

Also, the tags should be left-justified in `StoreDetailsScreen`, but when I tried to left justify everything, the widths would expand and it would look pretty gross. The widths of the chips only look right on this screen because of `alignItems: 'center'` for some reason. Any tips would be APPRECIATED! :'(

Tested tag overflow by making the widths comically large. Tested with edge case stores like Grubbs and Nam's.

### Screenshots
IMG_2714
IMG_2715

CC: @tommypoa  @wangannie


screens/map/StoreDetailsScreen.js
```diff
@@ -0,0 +1,227 @@
...
+                    )}
+                    style={styles.chip}>
+                    EBT
+                  </Chip>
+                )}
+                {wic && (
+                  <Chip
+                    icon={() => (
+                      <FontAwesome5
+                        name="heart"
```
wangannie: Add prop `solid` to make it the filled icon

wangannie:
Notes on what I changed:
* Modified sizing of the chips to better match the Figma (smaller cards, smaller font, text color, Poppins font)
* As a temporary fix for wrapping issues, I made the description text truncate into ellipses
* Some temporary copy changes ex 'Accepts Healthy Rewards' instead of 'Participates in Healthy Rewards' because long text is HARD AND WRAPPING IS SO ANNOYING AH
* Left-aligned chips in StoreDetails

components/store/StoreCard.js
```diff
@@ -55,89 +61,153 @@ export default function StoreCard({
...
+                  size={10}
+                  color={Colors.darkerGreen}
+                  style={{ marginTop: -10 }}
+                />
+              )}
+              textStyle={styles.chipText}
+              style={styles.chip}>
+              <Caption color={Colors.darkerGreen}>WIC</Caption>
+            </Chip>
+          )}
```
tommypoa, line 141: Can probably be a separate component since only the caption and font name is changed for similar components below!

tommypoa: Realised @wangannie also noted this in her review

PR #119, opened 2020-04-19
Filters for Store List

anniero98:

# What's new in this PR

### In sum
- Added filters for the store list: "Open now", "Products in stock", "EBT", "WIC", "Snap Match", and "Healthy Rewards Accepted".
- Partially waiting on finalized designs, but in its current state it is usable and conforms more-or-less

### Functionality
- Added state object `filters` with various property names that map directly to `store` object properties (except for the `openNow` property).
- Added horizontal scrollview of `Chip` components, selectable
- Implemented filtering so that filters stack, and search is also applied
- Realized that, in a similar manner to how we initially implemented Auth button permissions, we don't need to keep filtered stores in state. Rather, we can just calculate and update the list in the `render` function, since all the information we need is in state.

### Styling
- Updated the chips styling in `styled/store.js`. Most of the work was really done in #98, I just messed around with the heights and margins until things seemed alright.

- Instead of using the `Chip` `built-in` `selected` property, I implemented filter-selection with booleans, since we needed them in state to filter the stores correctly anyway. Also, I couldn't see a good way to switch on `filters[name]` for the styling, text color, _and_ icon color.

### Other details
- Bug fix: moved the formatting for `formatPhoneNumber` into `mapUtils/updateStoreData` into creation of a `phoneNumber` property for stores (when initializing the data)

- In `StoreDetailsScreen`, had to un-format it https://stackoverflow.com/questions/1862130/strip-all-non-numeric-characters-from-string-in-javascript

- Cleanup: created a `storeOpenStatus` property for stores in `mapUtils/updateStoreData` and called `computeOpenHours` to initialize it. This is a string; if it contains `Open` then store is currently open. Removed logic explicitly calling this function outside of `mapUtils`

- Added `productsInStock`, `snapOrEbtMatching`, `couponProgramPartner`, etc boolean fields.
**Airtable does not return fields which are empty / null from the API, so we need to initialize these ourselves if we don't want to run into errors where we expect record objects to contain certain properties**

### Online sources

Referenced this: https://medium.com/better-programming/creating-a-multi-filter-function-to-filter-out-multiple-attributes-javascript-react-rails-5aad8e272142 (Has a paywall, though)

### Related PRs

## How to review

Sanity-check that the stores that show up in the list are as expected. Luckily, #98 makes testing this really easy (': Search strings should also stack, as shown in screen recording. 

## Next steps

Styling: Designs aren't finalized, and though I more or less followed Figma for the initial pixel values the filter chips still seem kind of small.

Code: Pretty sure I should extract the filter chips to a component because the code is so repetitive it drives me a little crazy. At this moment in time I am not really sure how to though since they all have their own functions and icons etc... bit of a headache. No matter what we'll have to do `filters.<filter-type-name>` everywhere. But if we're in a rush to merge this we can refactor it later.

### Screenshots

image

https://www.loom.com/share/7e47a6ca711b4b139da5b6af4cc07097

CC: @wangannie

### DIFF
--- a/screens/map/StoreListScreen.js
+++ b/screens/map/StoreListScreen.js
@@ -2,9 +2,17 @@ import { FontAwesome5 } from '@expo/vector-icons';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { FlatList, ScrollView, View } from 'react-native';
-import { SearchBar } from 'react-native-elements'; // @tommypoa: Create styled-component for this
+import { SearchBar } from 'react-native-elements';
+// @tommypoa: Create styled-component for this
 import { Chip } from 'react-native-paper';
-import { Body, ButtonLabel, Caption, NavHeaderContainer, Title } from '../../components/BaseComponents';
+
+import {
+  Body,
+  ButtonLabel,
+  Caption,
+  NavHeaderContainer,
+  Title,
+} from '../../components/BaseComponents';
 import StoreCard from '../../components/store/StoreCard';
 import Colors from '../../constants/Colors';
 import { ColumnContainer, RowContainer } from '../../styled/shared';
@@ -107,11 +115,15 @@ export default class StoreListScreen extends React.Component {
                 />
               }
               inputStyle={styles.input}
+              // eslint-disable-next-line no-return-assign
               ref={search => (this.search = search)}
             />
           </ColumnContainer>
         </NavHeaderContainer>
-        <ScrollView horizontal style={{ height: 52 }}>
+        <ScrollView
+          horizontal
+          showsHorizontalScrollIndicator={false}
+          style={{ height: 52 }}>
           {/* Filter Chips */}
           <Chip
             icon={() => (
@@ -254,7 +266,9 @@ export default class StoreListScreen extends React.Component {
                 solid
                 size={10}
                 color={
-                  filters.rewardsAccepted ? Colors.lightest : Colors.darkerOrange
+                  filters.rewardsAccepted
+                    ? Colors.lightest
+                    : Colors.darkerOrange
                 }
                 style={{ marginTop: 0 }}
               />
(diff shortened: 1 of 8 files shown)


screens/map/StoreListScreen.js
```diff
@@ -94,8 +111,173 @@ export default class StoreListScreen extends React.Component {
             />
           </ColumnContainer>
         </NavHeaderContainer>
+        <ScrollView horizontal style={{ height: 52 }}>
```
wangannie, line 114: Hide the scrollbar

styled/store.js
```diff
@@ -107,23 +107,42 @@ export const styles = StyleSheet.create({
   input: {
     fontFamily: 'poppins-regular',
   },
-  chip: {
+  tagChip: {
     backgroundColor: Colors.lightestGreen,
     color: Colors.darkerGreen,
     marginRight: 6,
-    marginBottom: 4,
```
wangannie, line 114: Wait we need this! That's A & S looked super weird with 2 lines of tags

anniero98: wait I only removed it because it seems very odd to have both `marginVertical` and `marginBottom` and I expected that `marginVertical` overrode the latter - why wasn't it just `marginBottom` and `marginTop`?

wangannie: I had to set `marginVertical` to 0 to override the defaults https://github.com/callstack/react-native-paper/blob/master/src/components/Chip.tsx#L333-L338 because it wasn't shrinking to my custom height (see my message in #ask), and I needed it to have 0 top margin.
image


There's probably a cleaner way to do this but I'm not super sure what combination of these margins work best.

anniero98: ok re-ordered them for my sanity but if it works it works!

anniero98:
> also yes I will work on componentalizing the chips
@thumn I will try to get some advice!! I bet someone knows how to but anyway dw about it for now!

PR #128, opened 2020-04-23
Lint and prettify entire repo

anniero98:

# What's new in this PR

- Updated `.eslintrc.js`
- Disabled for specific lines where we're ignoring the linter's advice (for now, at least)
- Updated prettier config with `arrowParens: always` for similar reasons as enabling trailing commas
- No errors in repo (still some warnings for console logging but it's whatever)

### Related PRs

Analogous to https://github.com/calblueprint/dccentralkitchen-clerks/pull/50

## How to review

Make sure everything is still ok!

## Next steps

N/A

### Screenshots

N/A

CC: @anniero98 @wangannie


components/rewards/Transaction.js
```diff
@@ -40,9 +40,9 @@ function Transaction(props) {
             <Caption color={Colors.secondaryText}>
               {`${date.toLocaleDateString('en-US', options)} • ${storeName}`}
             </Caption>
-            <Subhead>{`$5 reward unlocked`}</Subhead>
+            <Subhead>$5 reward unlocked</Subhead>
```
wangannie, line 43: Oh wait are you switching these back? Or in a later PR

anniero98: I pulled from `/constants` in #129, but to be clear this pattern is correct as well if no JSX is used. ESLint just wants you to be clear so it enforces template strings as long as a string is being 'concatenated' with one or more variables

PR #129, opened 2020-04-23
Add participating stores to Rewards screen

anniero98:

# What's new in this PR

- Added the list of stores which accept rewards to the 'Healthy Rewards' tab.
- Clicking the store name redirects to the map (just as clicking the store card in `StoreListScreen` does)

### Unrelated fixes
- Pulled `rewardPointValue` and `rewardDollarValue` from `/constants` for the cards from #93 
- Added `key` for those lists (would happen rarely in PROD assuming customers have small total purchase amounts in $)
- Added testing bypass button back in `dev` mode (necessary to test this feature since you need to be able to see rewards)
- Renamed a function because it was no longer transitioning to `StoreDetails` LOL

### Online sources

We should fix this error probably? "never" is a scary word 🤷‍♀️ (lol)
https://www.lahteenlahti.com/fixing-virtualizedlists-should-never-be-nested-inside-plain-scrollviews/

### Related PRs

branches off #128 
update #93 

### Styling
Not sure if spacing still needs to be adjusted, especially the indent.

### Functionality
Log in, make sure nothing breaks, and that stores redirect correctly.

Try updating the `rewards accepted` value in the [DEV] Airtable **(but remember to switch it back after, please)**

If erroneous lint/prop-types errors show up, please review #128 

## Next steps

N/A 

- This won't scale well (i.e when > 5 stores are accepting rewards the design should probably be updated, or this feature removed entirely).

* Noticed some very edge-case bugs (should _never_ happen in PROD, though) unrelated to the current PR
* If transaction does not have a `rewardsUnlocked` or `rewardsApplied` value (or whichever) the `Array(rewardsUnlocked)...` throws an error like 'Array size too big', probably because the value is undefined

### Screenshots

image

https://www.loom.com/share/8f9166e49f9544edae54188634e6bd53

CC: @wangannie


components/rewards/RewardsHome.js
```diff
@@ -25,11 +27,31 @@ function createList(n) {
...
+        <Overline style={{ marginTop: 24, marginBottom: 4 }}>
+          Participating Stores
+        </Overline>
+        {participating.map((store) => {
+          return (
+            <Subhead
+              style={{ marginLeft: 8 }}
+              key={store.id}
+              onPress={() =>
+                navigation.navigate('Stores', {
```
wangannie, line 46: Hmm this is really really slow to load for some reason. Could we do first do a `navigation.goBack` then do `changeCurrentStore` from MapScreen to set the store if that's not too clunky?

wangannie: Just so it doesn't leave the user staring at the screen wondering what's happening/what's supposed to happen.

Also, can you wrap these in touchable opacities so there's some response when these are clicked? Even that might help indicate that some action is supposed to happen bc rn it's quite ambiguous.

anniero98: > Could we do first do a navigation.goBack then do changeCurrentStore from MapScreen to set the store if that's not too clunky?

AFAIK this isn't possible? The method that gets invoked is `componentWillReceiveProps`, which is setting the store in `MapScreen`

> can you wrap these in touchable opacities so there's some response when these are clicked

👍

components/rewards/RewardsHome.js
```diff
@@ -25,11 +34,32 @@ function createList(n) {
...
 
-function RewardsHome({ customer }) {
+function RewardsHome({ customer, participating }) {
+  const navigation = useNavigation();
   const rewardsAvailable = parseInt(customer.points, 10) / rewardPointValue;
   const pointsToNext = parseInt(customer.points, 10) % rewardPointValue;
   return (
     <ScrollView style={{ marginLeft: 16, paddingRight: 16 }}>
+      <ColumnContainer>
+        <Overline style={{ marginTop: 24, marginBottom: 4 }}>
```
wangannie, line 44: See the latest designs -- this is moved to the bottom (under Avail rewards).

Also, this is included in the rewards screen even in Guest Mode

anniero98:
Updated styling + added to guest mode
image
image

components/rewards/ParticipatingStores.js
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <ColumnContainer
+      style={guest ? { marginLeft: 16, marginBottom: 24 } : { marginTop: 28 }}>
+      <Overline style={{}}>Participating Stores</Overline>
+      {participating.map((store) => {
+        return (
+          <TouchableOpacity key={store.id}>
+            <Subhead
+              style={{ marginLeft: 12 }}
+              onPress={() =>
```
wangannie, line 19: When I click this, it doesn't show the map marker as selected.

PR #131, opened 2020-04-23
Add fix for heisenbug Airtable error in MapScreen

anniero98:

# What's new in this PR

- In `_populateInitialStoreProducts`, the call to `_populateStoreProducts` depends on `setState` for `store` completing. There's a race condition here that we caught w/ Ethan & Thu debugging but the bug is still live in PROD.
- Fixed by adding a null check
- Added (belated) logging to Sentry

### Related PRs

N/A

## How to review

Difficult to reproduce the bug in the first place 😅 

## Next steps

We should add logging to all the `try/catch` blocks and verify that we are handling all possible promise rejections + add `try/catch` and logging if not. That'll help with diagnosing the white-screen bug (which this may or may not fix)

### Screenshots

image

This should stop happening

CC: @wangannie

### DIFF
--- a/screens/map/MapScreen.js
+++ b/screens/map/MapScreen.js
@@ -121,13 +121,15 @@ export default class MapScreen extends React.Component {
   };
 
   _populateStoreProducts = async store => {
-    try {
-      const products = await getProductData(store);
-      if (products) {
-        this.setState({ storeProducts: products });
+    if (store) {
+      try {
+        const products = await getProductData(store);
+        if (products) {
+          this.setState({ storeProducts: products });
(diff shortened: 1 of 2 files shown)


screens/map/MapScreen.js
```diff
@@ -23,6 +23,7 @@ import {
   DragBar,
   SearchBar,
 } from '../../styled/store';
+import { logErrorToSentry } from '../lib/logUtils';
```
wangannie, line 26: Says this is missing -- must be `../../lib/logUtils`

PR #97, opened 2020-04-18
Phone Number Authentication + Forgot Password Functionality

kennethlien:
[//]: # "These comments are meant for your reference. They are invisible and don't need to be deleted!"

# What's new in this PR
It is now possible to reset your password by clicking 'forgot password' in the log in screen.
Signing up + Resetting passwords now sends a text to your phone after verifying you're not a robot.

Changes:
- `SignUpScreen.js` initializes firebase, no longer creates a new user on Airtable until the phone number is verified.
- `LogInScreen` has a new button, 'Forgot Password', to reset a user's password.
- To reset password, the `LogInScreen` redirects to the new `PasswordResetScreen`, where after verifying a user's number, they will be prompted to enter their new password and return to the log in screen.
Note: Both signing up and resetting password follow the same flow to verify, detailed below.

Verification Flow:
- From either `SignUpScreen.js` or `PasswordResetScreen`, `RecaptchaVerifierModal` is opened, and after being closed, `VerificationModal.js` will be opened.
- Props containing the state of the screen the verification modal was opened will be passed in, so that when the correct code is entered the modal will close and update the state of the screen.

There were some funky ways of passing in functions that eslint really didn't like, but it made callbacks work so they're there for now.

[//]: # "Describe what's new in this PR in a few lines. A description and bullet points for specifics will suffice."

## Relevant Links
https://docs.expo.io/versions/latest/sdk/firebase-recaptcha/
### Online sources
https://docs.expo.io/versions/latest/sdk/firebase-recaptcha/

[//]: # "Optional - copy links to any tutorial or documentation that was useful to you when working on this PR"

### Related PRs

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc; if this is a refactor, feel free to add PRs that previously modified this code"

Dependent on Firebase PR.

## How to review

[//]: # "The order in which to review files and what to expect when testing locally"

To prevent yourself from being logged out while testing, use 6505555056 as your number, and 666666 as your code.

Try to close out the recaptcha/create an account/reset passwords in any way that causes unexpected functionality.

Known issues are swiping down on the recaptcha will break the app and require clearing the app from memory, pressing close will require you to resend your code.

## Next steps

[//]: # "What's NOT in this PR, doesn't work yet, and/or still needs to be done"

## Tests Performed, Edge Cases

[//]: # "Hopefully we will add a testing suite/CI soon, but until then note down the steps you took to test locally"

### Screenshots

[...]

PR #142, opened 2020-05-01
Fixed keyboard covering on Log In and Sign Up

### LINKED ISSUE #63: Fix keyboard covering issues on Login and Signup
Replace the temporary fix of shrunken spacing on login and signup from #56 with keyboardAvoidingView or some long-term solution. With the current spacing, it can still be blocked by the keyboard (iPhone 8 on the left)

image

wangannie:

# What's new in this PR
* Resolves #63 where the keyboard covered the text fields and buttons on smaller screens.
* Added Keyboard Avoiding to new auth screens from #97 

Note: I was originally using a regular `KeyboardAvoidingView` for `AuthScreenContainer` but I was having issues where the content would end up overlapping since it just adjusted with `flex: 1` and refused to go off the screen. I decided to switch to wrapping everything in a `ScrollView` inside `KeyboardAvoidingView` so the spacing wouldn't have to be adjusted based on screen size, and it makes sure that all of the screen content can always be accessible. Unfortunately, this sacrifices the smoothness of a pure `KeyboardAvoidingView`, but is probably the best solution for accessibility across different types of devices.

5/5 Update: Turned all the repeated `ScrollView`s into `AuthScrollContainer` with all the needed attributes.

## How to review
* Go through the login and signup flows and see if the behavior is intuitive. You should be able to:
  * Tap a text field/button without having to dismiss the keyboard
  * Moving away from the first text field in both LogIn and SignUp should automatically scroll you to the bottom so the button should be visible.
  * You should always be able to scroll up to hit the back button
  * On larger devices where all of the screen content is visible, there shouldn't be any visible auto-scrolling behavior and everything should look/work like it used to.

* I was having issues with my emulator, so this still needs to be tested on Android @12aschen Use [this expo link ](https://exp.host/@wangannie/healthy-corners-rewards-androidtestkeyboard)to test on Android.

### Online sources
Shoutout to @averyyip and https://github.com/calblueprint/1951coffee/pull/13 for the inspiration in 1951 to do a `ScrollView` within a `KeyboardAvoidingView` I still thought their interaction was rather jank, so I added the automatic scrolling to the bottom to make sure the proper fields will show.

### Related PRs
Originally began updating this in #97 but decided to split it up into a separate PR.

## Next steps
n/a

### Screenshots
https://www.loom.com/share/0f09f45e55fa41d1b1ee4ddb0efa739f

Behavior in new auth screens
https://www.loom.com/share/b70aed83bbfa47d591553e111b51c57f

CC: @anniero98 


screens/auth/LogInScreen.js
```diff
@@ -160,55 +160,66 @@ export default class LogInScreen extends React.Component {
...
-          </FilledButtonContainer>
-
-          {/* TODO @tommypoa: Forgot password functionality
+        <ScrollView
+          ref={(ref) => {
+            this.scrollView = ref;
+          }}
+          style={{ flex: 1 }}
+          keyboardShouldPersistTaps="always"
+          showsVerticalScrollIndicator={false}>
```
kennethlien, line 169: idk what this is but i assume it's doing the magic thing, you're an absolute legend

screens/auth/SignUpScreen.js
```diff
@@ -300,21 +306,22 @@ export default class SignUpScreen extends React.Component {
...
 
             <AuthTextField
               fieldType="Phone Number"
               value={this.state.values[signUpFields.PHONENUM]}
-              onBlurCallback={(value) =>
-                this.updateError(value, signUpFields.PHONENUM)
-              }
+              onBlurCallback={(value) => {
+                this.updateError(value, signUpFields.PHONENUM);
+              }}
```
tommypoa, line 324: If for some reason the user scrolls up while typing the phone number, perhaps to change their name (such that the sign up button is hidden but the password field is not hidden), tapping on the password field does not animate to show the sign up button – maybe add the blur callback to the phone number field too? Definitely an edge case though

wangannie: Yes nice catch! I added it to both fields now.

Issue #143, opened 2020-05-01
Text misaligned in onboarding


wangannie:
![image](https://user-images.githubusercontent.com/21160510/80822663-0db8fc80-8b90-11ea-8346-a2c5f199c97d.png)

Noted by @anniero98 (iPhone X with accessibility large text) and @12aschen (Pixel 2, regular text)

PR #146, opened 2020-05-02
Bug Fix: Onboarding title text alignment

tommypoa:

# What's new in this PR

- Fix for issue #143 
- Text alignment property does not propagate from parent class – had to manually add inline styling 
## How to review

[//]: # 'The order in which to review files and what to expect when testing locally'
- Configure your fonts to be larger and see if the title aligns!

## Relevant Links
- #143 
## Next steps

- Possibly add textAlign property to base components for `Title` and/or `Body`, but only few instances in the entire app so not sure if it is too necessary

### Screenshots

CC: @wangannie

### DIFF
--- a/screens/auth/OnboardingScreen.js
+++ b/screens/auth/OnboardingScreen.js
@@ -1,7 +1,7 @@
 import { FontAwesome5 } from '@expo/vector-icons';
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Image } from 'react-native';
+import { Image, ScrollView } from 'react-native';
 import Carousel, { Pagination } from 'react-native-snap-carousel';
 import {
   Body,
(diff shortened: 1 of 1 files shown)


screens/auth/OnboardingScreen.js
```diff
@@ -78,12 +78,14 @@ export default class OnboardingScreen extends React.Component {
             />
           ))}
         </RowContainer>
-        <ColumnContainer style={{ alignItems: 'center' }}>
-          <Title>{item.title}</Title>
-          <Body style={{ marginTop: 12, textAlign: 'center' }}>
-            {item.body}
-          </Body>
-        </ColumnContainer>
+        <ScrollView>
```
wangannie, line 81: GREAT IDEA i love this

PR #149, opened 2020-05-05
Fixed Logout and Readded Rewards Footer

kennethlien:

# What's new in this PR
- Added await to all instances of logging out, sorry I didn't catch this earlier and you had a whole PR about it @thumn 
- Readded `RewardsFooter.js` back to `MapScreen.js`, it now matches Figma!
- Changed `RewardsFooter.js` to a class to contain customer points to the file so it wouldn't have to be passed as a prop for the sake of not making `MapScreen.js` longer.

## How to review
See if we still have a previous bug where the points wouldn't update after logging out and signing in with another account.

[//]: # 'The order in which to review files and what to expect when testing locally'

## Relevant Links
https://reactnative.dev/docs/asyncstorage.html
Was reading this when I realized
### Online sources

[//]: # 'Optional - copy links to any tutorial or documentation that was useful to you when working on this PR'

### Related PRs

## Next steps
I have an instance or two of new logout locations, gotta make sure to add await to those too.

### Screenshots

image Uploading 81046015...


wangannie:
Ah the footer still doesn't update for me. It only refreshes if I refresh expo.

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
-  customer: PropTypes.object,
-  isGuest: PropTypes.bool,
-};
+export default class RewardsFooter extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      customer: null,
+      transactions: [],
+      isLoading: true,
```
wangannie, line 18: Remove this (seems like it was left from the copy paste)

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
+      isGuest: false,
+    };
+  }
+  // Load customer record & transactions
```
wangannie, line 22: Remove this

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
+          <View
+            style={{
+              flexDirection: 'row',
+              alignItems: 'center',
+            }}>
+            <FontAwesome5 name="star" solid size={16} color={Colors.lightest} />
+            <Subhead style={{ paddingLeft: 8 }} color={Colors.lightest}>
+              {this.state.isGuest
+                ? 'Learn about Healthy Rewards >'
+                : `${this.state.customer.points} points`}
```
wangannie, line 62: @thumn mentioned a while ago that this could be really confusing since this displays total points, and the Rewards page shows progress toward the next reward instead. It should probably be `rewardPointValue - pointsToNext` like on the rewards progress bar in `RewardsHome`. I followed up with @12aschen for updated designs on this.

screens/map/MapScreen.js
```diff
@@ -373,9 +369,7 @@ export default class MapScreen extends React.Component {
             zIndex: 1000,
           }}
           onPress={() => this.props.navigation.navigate('RewardsOverlay')}>
-          <View>
-            <Subhead color="#fff"> Healthy Rewards </Subhead>
-          </View>
+          <RewardsFooter>Yes</RewardsFooter>
```
wangannie, line 372: Get rid of 'Yes' — this can be self-closing `<RewardsFooter />`

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
+    try {
+      const customer = await getCustomersById(customerId);
+      this.setState({
+        isGuest,
+        customer,
+        isLoading: false,
+      });
+    } catch (err) {
+      console.error(err);
+      logErrorToSentry({
```
wangannie, line 35: If you're including this, you need to import it from `logUtils`!

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
+            <FontAwesome5 name="star" solid size={16} color={Colors.lightest} />
+            <Subhead style={{ paddingLeft: 8 }} color={Colors.lightest}>
+              {this.state.isGuest
+                ? 'Learn about Healthy Rewards >'
+                : `${this.state.customer.points} points`}
+            </Subhead>
+          </View>
+        )}
+        {this.state.customer && !this.state.isGuest && (
+          <Subhead color={Colors.lightest}>Your next reward &gt;</Subhead>
```
thumn, line 67: Probably would be a good idea to leave a comment explaining why we have `&gt;` here for the future devs? :0

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,72 @@
...
 
-RewardsFooter.defaultProps = {
-  customer: null,
-  isGuest: true,
-};
-
-export default RewardsFooter;
+  render() {
+    return (
+      <View
```
thumn, line 45: This is something that I realized super late but in `shared.js` there's a component called `SpaceBetweenRowContainer` that applies `space-between` and `row` to a `View` for you! (I'm definitely guilty of not using this in multiple PRs...) Not an important change but just something to consider reviewing if you ever want to shorten code :- )

kennethlien:
footer should update after our fix in settings

components/rewards/RewardsFooter.js
kennethlien: oops..

components/rewards/RewardsFooter.js
kennethlien: changed!

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,77 @@
...
+            <Subhead style={{ paddingLeft: 8 }} color={Colors.lightest}>
+              {this.state.isGuest
+                ? 'Learn about Healthy Rewards >'
+                : `${this.state.rewards} rewards`}
+            </Subhead>
+          </View>
+        )}
+        {/* &gt; is a code for >, but since we can't use that, we use &gt; */}
+        {this.state.customer && !this.state.isGuest && (
+          <Subhead color={Colors.lightest}>View your rewards &gt;</Subhead>
```
wangannie, line 72: This should only show # rewards if they have rewards. It should show points if they don't have any. Check the latest Figma updates!

components/rewards/RewardsFooter.js
```diff
@@ -1,51 +1,77 @@
...
+          <View
+            style={{
+              flexDirection: 'row',
+              alignItems: 'center',
+            }}>
+            <FontAwesome5 name="star" solid size={16} color={Colors.lightest} />
+            <Subhead style={{ paddingLeft: 8 }} color={Colors.lightest}>
+              {this.state.isGuest
+                ? 'Learn about Healthy Rewards >'
+                : `${this.state.rewards} rewards`}
```
wangannie, line 66: Make sure this says 'reward' instead if they only have 1 reward.

PR #178, opened 2020-10-11
Favorite stores + delivery alerts

wangannie:

# What's new in this PR
Ability to 'favorite' stores to receive SMS alerts when the store receives new product deliveries.
- Added 2 new steps to onboarding process to select 'favorite' stores and enable SMS notifications
- Design updates on store drawer in map screen to highlight key info only
- Added notifications section to settings
- Expanded store search to support searching by zip code and store address
- Added integration with branch.io for deep linking support

Cleanup + various small fixes including:
- Added a `completeLogout` function to `AuthUtils` to reduce repeated code across logout functions
- Removed forced "Refresh to save changes" from Change Name screen

## How to review
NOTE: `StoreCardMini.js` is a temporary redesigned version of `StoreCard`

### Online sources

[//]: # 'Optional - copy links to any tutorial or documentation that was useful to you when working on this PR'

### Related PRs
n/a

## Next steps
- More efficient store search (currently it's a very simple filter) + sort results by relevance


lib/mapUtils.js
```diff
@@ -292,3 +302,70 @@ export function openDirections(latitude, longitude, storeName) {
...
+      favoriteStoreIds: favoriteStores,
+    });
+    success = true;
+  }
+  return success;
+}
+
+// Checks if a store is one of the customer's favorites
+// TODO @wangannie rename
+export async function isFavoritefromCustomer(storeId) {
```
wangannie, line 366: Not sure if this is an efficient way to check if a store is one of the customer's favorites. I thought it wouldn't make sense to add as a boolean to the `stores` list since that's loaded when the app opens, and I thought it'd get messy updating that through favoriting/unfavoriting.

navigation/stack_navigators/AuthStack.js
```diff
@@ -16,6 +18,8 @@ export default function AuthStackNavigator() {
       }}>
       <AuthStack.Screen name="Onboarding" component={OnboardingScreen} />
       <AuthStack.Screen name="PhoneNumber" component={PhoneNumberScreen} />
+      <AuthStack.Screen name="Permissions" component={PermissionsScreen} />
+      <AuthStack.Screen name="StoreSelect" component={StoreSelectScreen} />
```
wangannie, line 22: Not sure it would make more sense to include StoreSelect in StoresStack instead since this screen should still be usable after the initial onboarding process (for modifying selections later on)

screens/onboarding/PermissionsScreen.js
```diff
@@ -0,0 +1,262 @@
...
+      step: 1,
+    };
+  }
+
+  async getLocation() {
+    await this._findCurrentLocation();
+    this.navigateStoreSelect();
+  }
+
+  _findCurrentLocation = async () => {
```
wangannie, line 50: Many of the things in this screen were copied from `MapScreen` including `_findCurrentLocation`, `_populateInitialStoresProducts`, `_orderStoresByDistance`, etc. Would it make more sense to extract all of it out to a utils file or something like that? 

Also, it feels really inefficient that all of this is happening in `PermissionsScreen` and then AGAIN when the onboarding steps are done/app opens to the map, but also I'm not quite sure how I could eliminate that repetition/somehow pass the resulting ordered store list through

screens/onboarding/PermissionsScreen.js
```diff
@@ -0,0 +1,262 @@
...
+          ? region
+          : prevState.region,
+    }));
+    Analytics.setUserProperty(
+      'closest_store',
+      this.state.showDefaultStore ? 'default' : this.state.store.storeName
+    );
+  };
+
+  updateStep = () => {
```
wangannie, line 134: I wanted to consolidate the 2 permissions screens (selecting favorite stores and enabling notifications) into this 1 screen, so just have a `step` variable in the state + this callback function to update it (since completing step 1 takes you back to this page). 
1. Is this a good way to handle it? It feels very jank
2. It makes the navigation look a bit funky because the animation after finishing the 'select stores' step looks like it's going back, though it should look like it's going forward.

screens/onboarding/PermissionsScreen.js
```diff
@@ -0,0 +1,262 @@
...
+      });
+    }
+  }
+
+  async navigateStoreSelect() {
+    await this._populateInitialStores();
+    this.props.navigation.navigate('StoreSelect', {
+      navigation: this.props.navigation,
+      showDefaultStore: this.state.showDefaultStore,
+      stores: this.state.stores,
```
wangannie, line 187: I get a "Non-serializable values were found in the navigation state" which isn't a huge issue, but having to pass in the list of stores makes the StoreSelect screen much harder to re-use. Also the # of params seems excessive...

screens/onboarding/PermissionsScreen.js
ethanlee16: > Would it make more sense to extract all of it out to a utils file or something like that?

I think it would! You may find that it's not super convenient to decouple the statefulness from these functions, which is why Hooks attempt to solve this problem by allowing you to reuse stateful/effectful logic across components 😄 As a hook, `_findCurrentLocation` may look something like:

```tsx
function useCurrentLocation() {
  const [error, setError] = useState(null);
  const [region, setRegion] = useState(null);

  useEffect(() => {
    async getCurrentLocation() {
      try {
        const { status } = await Permissions.askAsync(Permissions.LOCATION);
        if (status !== 'granted') {
          Analytics.setUserProperty('location_permissions', 'denied');
          throw new Error('Permission to access location was denied');
        } else {
          const location = await Location.getCurrentPositionAsync({
            timeout: 3000,
          });
          const region = {
            latitude: location.coords.latitude - deltas.latitudeDelta / 3.5,
            longitude: location.coords.longitude,
            ...deltas,
          };
          setRegion(region);
          Analytics.setUserProperty('location_permissions', 'granted');
        }
      } catch (err) {
        Analytics.setUserProperty('location_permissions', 'error');
        logErrorToSentry({
          screen: 'MapScreen',
          function: 'useCurrentLocation',
          error: err,
        });
        setError(err.message);
      }
    }
    getCurrentLocation();
  });

  return { error, region };
}
```

Of course, Hooks won't quite work with class components. But maybe good as food for thought 😄

screens/onboarding/StoreSelectScreen.js
```diff
@@ -0,0 +1,204 @@
...
+      } else {
+        selectedStores = [...state.selectedStores, storeId];
+      }
+      return {
+        selectedStores,
+      };
+    });
+  };
+
+  filterStore = (searchStr) => {
```
wangannie, line 62: There's quite a bit of repetition between this file and the `StoreListScreen`, but does that warrant moving this out to a `utils` file?

screens/settings/NotificationsScreen.js
```diff
@@ -0,0 +1,171 @@
...
+import Colors from '../../constants/Colors';
+import { getCustomersById, updateCustomers } from '../../lib/airtable/request';
+import { logErrorToSentry } from '../../lib/logUtils';
+import { ContentContainer, ResourceItemCard } from '../../styled/resources';
+
+export default function NotificationsScreen(props) {
+  const { navigation } = props;
+  const [notifications, setNotifications] = useState(false);
+  const [deliveryAlerts, setDeliveryAlerts] = React.useState(false);
+  const [notifsChanged, setNotifsChanged] = useState(false);
```
wangannie, line 27: I have these 'changed' variables to check if edits have been made so that the save button will pop up/disappear only if changes have been made, + it will show a warning if you try to exit without saving changes. Now that I think about it, that kind of feels like overkill.... so is this a bad thing to do?

PR #184, opened 2020-12-23
Refactor MapScreen to functional component

wangannie:

# What's new in this PR

- MapScreen is finally a functional component so we can save a lot of trouble and use HOOKS
  -  No more deprecated `componentWillReceiveProps` !!
  - Lots of helper functions are extracted out to `mapUtils` or converted to custom hooks so they can be reused in #178
  - Significant loading speed/UX improvements, especially with the addition of a loading screen (currently a placeholder to be redesigned).
 
**The new MapScreen flow:**
1. `useCurrentLocation` prompts the user for location access and if granted, it sets the user's location in`currentLocation` (otherwise `null`). Location permission status is saved in `locationPermissions` as either 'granted', 'denied', or 'error'
2. `useStores` loads the list of stores from Airtable, not in any sorted order.
3. Store distance calculations and sorting happen directly since they are neither async nor expensive. The sorted `store` list (with the added `distance` attribute) is passed into `StoreListScreen` and to other store-related components that reference distance.
4. Once stores and location permissions are loaded, a `useEffect` sets the initial `currentStore` to focus on.
5. `useStoreProducts` loads the products for `currentStore`, refreshing whenever `currentStore` is updated.

**Other misc. changes**
- `findStoreDistance` returns null if the store distance is greater than 100 miles (NPO said if the user is >100 miles away, keep the focus in the DC area and treat it as if their location settings are off). This eliminated the need for other things like `showDefaultStore` and `seeDistance` being passed around to show/hide store distances since any non-null `distance` values should be displayed.
- Converted `StoreListScreen` to functional component
- Moved `ButtonContainer` that wrapped around the `RewardsFooter` into the actual `RewardsFooter` component
- Cleaned up incorrectly used state variables in `RewardsFooter` and `SettingsScreen`
## How to review
- I'll leave some comments in the actual code on particular areas I'm confused about, but in general, I'd appreciate feedback on:
  - general best practice with hooks -- when/where/how they should be used
  - ways I could clean up my code, naming conventions, etc.

### Online sources
https://reactjs.org/docs/hooks-custom.html
https://www.robinwieruch.de/react-hooks-fetch-data
https://reactjs.org/docs/thinking-in-react.html

## Next steps
- Integrate changes with store favoriting + delivery alerts #178 

### Screenshots
Here's how it loads AFTER this refactor. SO MUCH FASTER!
https://www.loom.com/share/090a25122bcd44a09e5960f2893b083e

### DIFF
--- a/lib/mapUtils.js
+++ b/lib/mapUtils.js
@@ -317,8 +317,8 @@ export function findStoreDistance(region, store) {
   return null;
 }
 
-export function sortByDistance(region, a, b) {
-  return findStoreDistance(region, a) - findStoreDistance(region, b);
+export function sortByDistance(a, b) {
+  return a.distance - b.distance;
 }
 
 export function findDefaultStore(stores) {
--- a/lib/mapUtils.js
+++ b/lib/mapUtils.js
@@ -383,10 +383,7 @@ export function useStores() {
       setStores(storeData);
     } catch (err) {
       console.error('(useStores) Airtable:', err);
-      logErrorToSentry({
-        function: 'useStores',
-        error: err,
-      });
+      logErrorToSentry({ function: 'useStores', error: err });
     }
   };
   useEffect(() => {
@@ -420,4 +417,36 @@ export function useStoreProducts(currentStore) {
     populateStoreProducts(currentStore);
   }, [currentStore]);
   return storeProducts;
+}
+
+/**
+ * Filters stores based on set of filters and/or a search string.
+ */
+export function useFilteredStores(stores, filters, searchStr) {
+  const [filteredStores, setFilteredStores] = useState(stores);
+  const filterStores = () => {
+    if (stores) {
+      const selectedFilters = Object.keys(filters).filter(
+        (key) => filters[key]
+      );
+      setFilteredStores(
+        stores.filter(
+          (store) =>
+            store.storeName.toLowerCase().includes(searchStr.toLowerCase()) &&
+            (selectedFilters.length === 0 ||
+              selectedFilters.every((name) => {
+                // 'Open Now' is not a boolean property that exists
+                if (name === 'openNow') {
+                  return store.storeOpenStatus.includes('Open');
+                }
+                return store[name];
+              }))
+        )
+      );
+    }
+  };
+  useEffect(() => {
+    filterStores(filters);
+  }, [stores, filters, searchStr]);
+  return filteredStores;
 }
--- a/lib/mapUtils.js
+++ b/lib/mapUtils.js
@@ -1,4 +1,6 @@
 import * as Analytics from 'expo-firebase-analytics';
+import convertDistance from 'geolib/es/convertDistance';
+import getDistance from 'geolib/es/getDistance';
 import { Alert, Clipboard } from 'react-native';
 import { showLocation } from 'react-native-map-link';
 import {
@@ -8,9 +10,15 @@ import {
   today,
 } from '../constants/DaysOfTheWeek';
 import { googlePlaceId } from '../constants/Map';
+import RecordIds from '../constants/RecordIds';
 import { getAllStores, getProductsByIds } from './airtable/request';
 import { formatPhoneNumber } from './authUtils';
 import { logErrorToSentry } from './logUtils';
+
+export const deltas = {
+  latitudeDelta: 0.01,
+  longitudeDelta: 0.01,
+};
 
 // Takes in a string input hours and outputs a dictionary mapping
 // each day of the week to the store hours for that day.
@@ -291,4 +299,40 @@ export function openDirections(latitude, longitude, storeName) {
     googlePlaceId,
     alwaysIncludeGoogle: true,
   });
+}
+
+export function findStoreDistance(region, store) {
+  const distanceMeters = getDistance(region, store);
+  // Convert distance to 'x.xx' form, in miles units
+  const distance = convertDistance(distanceMeters, 'mi').toFixed(2);
+  return distance;
+}
+
+export async function orderStoresByDistance(region, inputStores) {
+  console.log('IN order stores by dist mapUtils helper');
+  const sortedStores = [];
+  // Display distance in the StoreList
+  inputStores.forEach((store) => {
+    const currStore = store;
+    currStore.distance = findStoreDistance(region, store);
+    sortedStores.push(currStore);
+  });
+  // sorts in place
+  sortedStores.sort(function compare(a, b) {
+    return a.distance - b.distance;
+  });
+  return sortedStores;
+}
+
+export function findDefaultStore(stores) {
+  const defaultStore = stores.find((store) => {
+    return store.id === RecordIds.defaultStoreId;
+  });
+
+  const defaultRegion = {
+    latitude: defaultStore.latitude,
+    longitude: defaultStore.longitude,
+    ...deltas,
+  };
+  return { defaultStore, defaultRegion };
 }
--- a/lib/mapUtils.js
+++ b/lib/mapUtils.js
@@ -12,16 +12,11 @@ import {
   daysOfTheWeekFull,
   today,
 } from '../constants/DaysOfTheWeek';
-import { googlePlaceId, initialRegion } from '../constants/Map';
+import { deltas, googlePlaceId } from '../constants/Map';
 import RecordIds from '../constants/RecordIds';
 import { getAllStores, getProductsByIds } from './airtable/request';
 import { formatPhoneNumber } from './authUtils';
 import { logErrorToSentry } from './logUtils';
-
-export const deltas = {
-  latitudeDelta: 0.01,
-  longitudeDelta: 0.01,
-};
 
 // Takes in a string input hours and outputs a dictionary mapping
 // each day of the week to the store hours for that day.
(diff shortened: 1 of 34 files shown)


lib/mapUtils.js
```diff
@@ -292,3 +298,131 @@ export function openDirections(latitude, longitude, storeName) {
...
+/**
+ * If location permissions are granted and location is found, order stores by distance.
+ */
+export function useSortedStores(
+  locationPermissions,
+  stores,
+  setStores,
+  isSorted,
+  setSorted,
+  initialLocation
```
wangannie, line 403: This is one of the situations where there are SO MANY dependencies! This feels super excessive, but I _think_ I need these to make sure location/stores are fully loaded before attempting to sort.

`orderLoadedStores` only runs if:
- location permissions are granted
- stores are loaded
- stores haven't already been sorted
- user's location has been determined `initialLocation` is not `null`

wangannie: Also, I used `isSorted` to avoid this repeating forever, but it seems weird to keep since it's only purpose is to say that this function has successfully completed. Are there better ways to handle that?

screens/map/MapScreen.js
```diff
@@ -35,198 +36,158 @@ import {
...
-    const { status } = await Permissions.askAsync(Permissions.LOCATION);
-    // Error message not checked anywhere
-    if (status !== 'granted') {
-      this.setState({
-        locationErrorMsg: 'Permission to access location was denied',
-      });
-      Analytics.setUserProperty('location_permissions', 'denied');
-    } else {
+  // Only populate initial store products on first load
+  const initialLoadComplete = useRef(false);
```
wangannie, line 63: Some Medium article/forum told me to use a ref like this to make sure a `useEffect` only runs once. Would you say this is a common accepted way to do this?

lib/mapUtils.js
```diff
@@ -292,3 +298,131 @@ export function openDirections(latitude, longitude, storeName) {
...
+
+  const defaultRegion = {
+    latitude: defaultStore.latitude,
+    longitude: defaultStore.longitude,
+    ...deltas,
+  };
+  return { defaultStore, defaultRegion };
+}
+
+export function useCurrentLocation(setInitialLocation, setLocationPermissions) {
```
ethanlee16, line 337: Instead of passing through these state setters as callbacks, why not include these as part of the hook itself? You're already asking for location permissions in the hook, so I think you should consider having the hook return:

```javascript
const { loading, location, permissionStatus } = useCurrentLocation();
```

You want the statefulness of the location/location permissions status to be coupled with the hook itself (so you can extract it all instead of creating new stateful values every time you use the hook), unless I'm missing something?

lib/mapUtils.js
```diff
@@ -292,3 +298,131 @@ export function openDirections(latitude, longitude, storeName) {
...
+  useEffect(() => {
+    const getCurrentLocation = async () => {
```
ethanlee16, line 339: Nit: to flatten, you could move this outside of the effect and only call `getCurrentLocation` within the effect.

lib/mapUtils.js
```diff
@@ -292,3 +298,131 @@ export function openDirections(latitude, longitude, storeName) {
...
 }
+
+export function findStoreDistance(region, store) {
+  const distanceMeters = getDistance(region, store);
+  // Convert distance to 'x.xx' form, in miles units
+  const distance = convertDistance(distanceMeters, 'mi').toFixed(2);
+  return distance;
+}
+
+export async function orderStoresByDistance(region, inputStores) {
```
ethanlee16, line 309: Why is this async?

lib/mapUtils.js
ethanlee16: I would say that this could be a non-hook function that gets passed as a `.sort` callback once the stores have been loaded — this sorting seems like more of a transformation that can be done on-the-fly instead of updated in a stateful value.

That way you could use it like:

```javascript
const { loading, stores } = useStores(); // Or however you load in stores
if (loading) {
  // Return loading state
}

return stores.sort(orderByLocation).map((store) => <Store {...} />);
```

If the sorting function was costly or actually async (in my below comment, I called out that it might not have to be designated as async) — you could consider memoizing (with `useMemo`) to only run this transformation if the `stores` value changes.

screens/map/MapScreen.js
```diff
@@ -35,198 +36,158 @@ import {
...
-    await this._populateInitialStoresProducts();
-  }
+export default function MapScreen(props) {
+  const [locationPermissions, setLocationPermissions] = useState(null);
+  const [region, setRegion] = useState(initialRegion);
+  const [stores, setStores] = useState([]);
+  const [initialLocation, setInitialLocation] = useState(null);
+  const [currentStore, setCurrentStore] = useState(null);
+  const [storeProducts, setStoreProducts] = useState([]);
+  const [showDefaultStore, setShowDefaultStore] = useState(true);
```
ethanlee16, line 46: Your goal should be to be able to knock all of these hooks out by composing them with your custom hook 😄 Give that a shot and see how it feels.

screens/map/MapScreen.js
```diff
@@ -35,198 +36,158 @@ import {
...
+  const bottomSheetRef = React.useRef(null);
+  const mapRef = React.useRef(null);
```
ethanlee16, line 48: Nit: you pull out `useRef` already from the React module, invoke it directly.

ethanlee16 (commented):
Some initial comments on hook structure. Most of it relates to composing the statefulness you rely on in those hooks _into_ the hook definitions themselves. Let me know what you think!

screens/map/MapScreen.js
ethanlee16: I'm not sure I have full context into "only runs once" — is there a limitation of an effect with an empty set of deps (`[]`)  that you've found? Otherwise that would be the option that I would use.

lib/mapUtils.js
wangannie: It definitely didn't need to be 😅

screens/map/MapScreen.js
wangannie: I didn't go with the empty set of dependencies because I needed this to run until my initial conditions passed (locationPermissions set, stores loaded). I ended up using this workaround to make it keep checking if the conditions were met, then execute `populateInitialStoreProducts` once.

screens/map/MapScreen.js
wangannie: When you say 'all' do you really mean *all*?  I knocked out most of them, but I'm still left with `useState` for `currentStore` and `region` (which is the user's location panning/zooming around on the map). It still makes sense to keep these right?

screens/map/MapScreen.js
```diff
@@ -35,198 +35,100 @@ import {
...
+  const storeProducts = useStoreProducts(currentStore);
+  const { locationPermissions, initialLocation } = useCurrentLocation();
+  let stores = useStores();
 
-  async componentDidMount() {
-    // We get current location first, since we need to use the lat/lon found in _populateIntitialStoresProducts
-    await this._findCurrentLocation();
-    await this._populateInitialStoresProducts();
+  if (locationPermissions === 'granted') {
+    stores = orderStoresByDistance(initialLocation, stores);
```
wangannie, line 49: Ok I think I'm still a little confused about how I should handle this. Like you mentioned, `orderStoresByDistance` is neither expensive nor async, and my initial `useSortedStores` mess was totally unnecessary. Still, I don't think I fully understood what you meant in [your comment from earlier](https://github.com/calblueprint/dccentralkitchen/pull/184#discussion_r549527181) about turning this into a transformation that can be done on-the-fly. Where/when/how would I do that? 😅 

Also, even if it isn't expensive/async, do you think it'd still be a good idea to use `useMemo` just to prevent this from running on every render? I realized this is modifying `stores` every time it runs (though I'm not sure why since the order/distances shouldn't be changing), which means it's also triggering the `useEffect` below which has `stores` as a dependency, even though that should only be executed once.

screens/map/MapScreen.js
```diff
@@ -35,198 +35,100 @@ import {
...
-          function: '_findCurrentLocation',
-          error: err,
-        });
+  // This should only happen once on first load.
+  useEffect(() => {
+    if (
+      (!currentStore || !region) &&
+      locationPermissions &&
+      stores.length > 0
+    ) {
```
wangannie, line 62: This `useEffect` basically just sets the current store and region (map location) to defaults if the user has location disabled OR is over 100 miles away from the closest store.

Basically, it should only run:
- on the first load (`(!currentStore || !region)`)
- once locationPermissions has been set by `useCurrentLocation`
- once stores are loaded from `useStores`

Subsequent changing of current store/region is handled separately in `changeCurrentStore`.

The way I'm doing this feels really inelegant, but I'm struggling to figure out a better way to handle this. Thoughts?

lib/mapUtils.js
wangannie: I _think_ I fixed this now? But wow as soon as I saw this something finally clicked and I realized I had this massive gap in my understanding of hooks :'))

screens/map/MapScreen.js
ethanlee16: Yes, you're right — makes sense to keep `currentStore` and `region` since they're component-specific!

screens/map/MapScreen.js
ethanlee16: You are doing it! On-the-fly meaning somewhere in the render path (or right before it) — just not within an effect. I see that you need the ordered stores within your effect below, though. Since `orderStoresByDistance` returns a new array every time (which is fine), I think memoizing as an optimization to prevent unexpected updates could be a good idea.

I would also consider not overshadowing the `stores` name with the ordered stores value to aid own debugging with the changing values 😄 

> though I'm not sure why since the order/distances shouldn't be changing

Changing values in React always use a "shallow equal" comparison (does it pass `===`?), so new arrays with identical contents will be considered different (i.e., `[] === [] -> false`)

screens/map/MapScreen.js
ethanlee16: What if instead of using these conditions to set the `currentStore` and `region` with values in state, you allowed them to be empty/null and performed this handling as a "fallback" for those values? Would that be possible?

